PR #5, opened 2016-09-06
add diffy

nevillelyh:


andrewsmartin:
@nevillelyh What about template `diff` method with abstract methods for `getName` and such for the Avro, Protobuf etc subclasses? Just to have one place where the logic would change instead of 3.

nevillelyh:
Too many differences, not gonna work.

andrewsmartin:
Guess you're right, even the types matched on are different. +1

PR #12, opened 2017-03-09
wip sample

ravwojdyla:
This it RFC. @nevillelyh @andrewsmartin  (please give feedback when you have some free time).

This is suppose to cover basic use cases of sampling for pipeline testing.

The overall concept would be:
 * BigSampler is a "generic enough" sampling job capable of sampling Avro/BigQuery
 * sampling happens on user defined key(s) (either via avro-schema or fields specified CLI)

Example of execution:

```
> sbt pack
> ./target/pack/bin/big-sampler --sample=5  --input=<table-reference-inpute> --output=<table-reference-sample> --project=<project> --zone=<zone> --stagingLocation=<staging> --fields=user_id --runner=DataflowPipelineRunner
```

Manual flow for sampling and running a job on sampled data would require to sample on each input, but this process can be automated with orchestrator (example luigi).

### DIFF
--- a/src/main/scala/com/spotify/ratatool/io/FileStorage.scala
+++ b/src/main/scala/com/spotify/ratatool/io/FileStorage.scala
@@ -165,7 +165,7 @@ private class LocalStorage(protected val path: String)  extends FileStorage {
       (new File("."), new WildcardFileFilter(path))
     }
 
-    if (dir.exists()) {
+    if (dir.isDirectory) {
       FileUtils
         .listFiles(dir, filter, null)
         .asScala
--- a/src/main/scala/com/spotify/ratatool/io/FileStorage.scala
+++ b/src/main/scala/com/spotify/ratatool/io/FileStorage.scala
@@ -1,0 +1,186 @@
+/*
+ * Copyright 2016 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.ratatool.io
+
+import java.io._
+import java.net.URI
+import java.nio.ByteBuffer
+import java.nio.channels.Channels
+import java.nio.file.{Files, Path}
+import java.util.Collections
+import java.util.regex.Pattern
+
+import com.google.cloud.dataflow.sdk.options.PipelineOptionsFactory
+import com.google.cloud.dataflow.sdk.util.GcsUtil.GcsUtilFactory
+import com.google.cloud.dataflow.sdk.util.gcsfs.GcsPath
+import org.apache.avro.file.{SeekableFileInput, SeekableInput}
+import org.apache.commons.io.FileUtils
+import org.apache.commons.io.filefilter.WildcardFileFilter
+
+import scala.collection.JavaConverters._
+import scala.util.Try
+
+private[ratatool] object FileStorage {
+  def apply(path: String): FileStorage =
+    if (isGcsUri(new URI(path))) new GcsStorage(path) else new LocalStorage(path)
+  def isLocalUri(uri: URI): Boolean = uri.getScheme == null || uri.getScheme == "file"
+  def isGcsUri(uri: URI): Boolean = uri.getScheme == "gs"
+  def isHdfsUri(uri: URI): Boolean = uri.getScheme == "hdfs"
+}
+
+private[ratatool] trait FileStorage {
+
+  protected val path: String
+
+  protected def listFiles: Seq[Path]
+
+  protected def getObjectInputStream(path: Path): InputStream
+
+  def getAvroSeekableInput: SeekableInput
+
+  def exists: Boolean
+
+  def isDone: Boolean = {
+    val partPattern = "([0-9]{5})-of-([0-9]{5})".r
+    val paths = listFiles
+    val nums = paths.flatMap { p =>
+      val m = partPattern.findAllIn(p.toString)
+      if (m.hasNext) {
+        Some(m.group(1).toInt, m.group(2).toInt)
+      } else {
+        None
+      }
+    }
+
+    if (paths.isEmpty) {
+      // empty list
+      false
+    } else if (nums.nonEmpty) {
+      // found xxxxx-of-yyyyy pattern
+      val parts = nums.map(_._1).sorted
+      val total = nums.map(_._2).toSet
+      paths.size == nums.size &&  // all paths matched
+        total.size == 1 && total.head == parts.size &&  // yyyyy part
+        parts.head == 0 && parts.last + 1 == parts.size // xxxxx part
+    } else {
+      true
+    }
+  }
+
+  private def getDirectoryInputStream(path: String,
+                                      wrapperFn: InputStream => InputStream = identity)
+  : InputStream = {
+    val inputs = listFiles.map(getObjectInputStream).map(wrapperFn).asJava
+    new SequenceInputStream(Collections.enumeration(inputs))
+  }
+
+}
+
+private class GcsStorage(protected val path: String) extends FileStorage {
+
+  private val uri = new URI(path)
+  require(FileStorage.isGcsUri(uri), s"Not a GCS path: $path")
+
+  private lazy val gcs = new GcsUtilFactory().create(PipelineOptionsFactory.create())
+
+  private val GLOB_PREFIX = Pattern.compile("(?<PREFIX>[^\\[*?]*)[\\[*?].*")
+
+  override protected def listFiles: Seq[Path] = {
+    if (GLOB_PREFIX.matcher(path).matches()) {
+      gcs
+        .expand(GcsPath.fromUri(uri))
+        .asScala
+    } else {
+      // not a glob, GcsUtil may return non-existent files
+      val p = GcsPath.fromUri(path)
+      if (Try(gcs.fileSize(p)).isSuccess) Seq(p) else Seq.empty
+    }
+  }
+
+  override protected def getObjectInputStream(path: Path): InputStream =
+    Channels.newInputStream(gcs.open(GcsPath.fromUri(path.toUri)))
+
+  override def getAvroSeekableInput: SeekableInput =
+    new SeekableInput {
+      private val in = gcs.open(GcsPath.fromUri(path))
+
+      override def tell(): Long = in.position()
+
+      override def length(): Long = in.size()
+
+      override def seek(p: Long): Unit = in.position(p)
+
+      override def read(b: Array[Byte], off: Int, len: Int): Int =
+        in.read(ByteBuffer.wrap(b, off, len))
+
+      override def close(): Unit = in.close()
+    }
+
+  override def exists: Boolean = {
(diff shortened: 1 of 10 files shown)


ravwojdyla:
ping @nevillelyh @andrewsmartin any poc comments?

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+          diceElement(e, hash, s.toInt, n)
+        }
+        .saveAsAvroFile(output, schema = schema)
+      sc.close()
+      r
+    }
+  }
+
+  @tailrec
+  private def samplePctToInt(samplePct: Float, n: Int): (Float, Int) = {
```
andrewsmartin: This will (in many cases) produce an approximate sample size, right? Is that the intention? For example `samplePctToInt(0.1999, 100)` would be `(1.999, 1000)` and later gets floored down to `(1, 1000)` right? I guess at most this will be off by one, so this is 👍

src/main/scala/com/spotify/ratatool/io/FileStorage.scala
```diff
@@ -0,0 +1,186 @@
...
+
+private[ratatool] trait FileStorage {
+
+  protected val path: String
+
+  protected def listFiles: Seq[Path]
+
+  protected def getObjectInputStream(path: Path): InputStream
+
+  def getAvroSeekableInput: SeekableInput
```
nevillelyh: This is not really avro specific?

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+import scala.util.{Failure, Success, Try}
+
+
+object BigSampler {
+
+  private val log = LoggerFactory.getLogger(BigSampler.getClass)
+  private val utf8Charset = Charset.forName("UTF-8")
+
+  private val kissHashFun = Hashing.murmur3_128(System.currentTimeMillis.toInt)
+  private def kissHashFun(seed: Option[Int]): HashFunction = {
```
nevillelyh: `seed: Option[Int] = None` here and move the `val` into `case None` branch?

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+  private case class PathsAndSchemas(input: URI,
+                                     output: URI,
+                                     schema: Schema,
+                                     sampleSchema: Class[_])
+
+  // scalastyle:off cyclomatic.complexity
+  private def hashAvroField(r: GenericRecord, fields: Seq[Field], hasher: Hasher): Hasher = {
+    fields.foldLeft(hasher) { case (h, field) =>
+      //TODO: is there need to check for nulls in v?
+      val v = r.get(field.name())
```
nevillelyh: Yes should probably check `null`s since avro allows them and they're everywhere.

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+        hasher
+      } else {
+        val v = r.get(fieldParts.head)
+        val field = fieldOpt.get
+        if (field.getMode == "REPEATED") {
+          //TODO: support repeated fields
+          throw new UnsupportedOperationException(
+            "Repeated field is not supported as sampling key!")
+        }
+        field.getType match {
```
nevillelyh: Double check the BQ data type spec. IIRC one of `INTEGER` or `FLOAT` is not really `Long` or `Double` but a `String` because of JSON. So `asInstanceOf` will fail.
Also `v` might be `null` if the field is NULLABLE.

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+import scala.language.existentials
+import scala.util.{Failure, Success, Try}
+
+
+object BigSampler {
+
+  private val log = LoggerFactory.getLogger(BigSampler.getClass)
+  private val utf8Charset = Charset.forName("UTF-8")
+
+  private val kissHashFun = Hashing.murmur3_128(System.currentTimeMillis.toInt)
```
andrewsmartin: What kind of distribution does this result in? I don't know much about this hashing algo, would it be "reasonably" uniform for most key types?

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+          throw new UnsupportedOperationException(
+            "Repeated field is not supported as sampling key!")
+        }
+        field.getType match {
+          case "BOOLEAN" => h.putBoolean(v.asInstanceOf[Boolean])
+          case "INTEGER" => h.putLong(v.asInstanceOf[Long])
+          case "FLOAT" => h.putFloat(v.asInstanceOf[Float])
+          case "STRING" => h.putString(v.asInstanceOf[String], utf8Charset)
+          case "BYTES" => h.putBytes(v.asInstanceOf[ByteString].toByteArray)
+          case "TIMESTAMP" =>
```
nevillelyh: You can do `case "TIMESTAMP" | "DATE" ... => throw new UOE(s"${field.getType} ..."`

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+      log.info(s"Will sample from BigQuery table: ${inputTbl}, output will be ${outputTbl}")
+
+      val schema = patchedBigQueryService.getTable(inputTbl).getSchema
+
+      // defeat closure TableSchema and TableFieldSchema are not serializable
+      val schemaStr = JsonSerDe.toJsonString(schema)
+      @transient lazy val schemaFields =
+        JsonSerDe.fromJsonString(schemaStr, classOf[TableSchema]).getFields.asScala.toList
+      val (s, n) = samplePctToInt(samplePct, 100)
+      val r = sc.bigQueryTable(inputTbl)
```
nevillelyh: Is it possible to delegate the hash to BQ? How much more work is it?

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+      val inputTbl = parseAsBigQueryTable(input).get
+      val outputTbl = parseAsBigQueryTable(output).get
+      sampleBigQueryTable(sc, inputTbl, outputTbl, fields, samplePct, seed.map(_.toInt))
+    } else if (parseAsURI(input).isDefined) {
+      // right now only support for avro
+      require(parseAsURI(output).isDefined,
+        s"Input is a URI: `${input}`, output should be a URI too, but instead it's `${output}`.")
+      require(fields.isEmpty && sampleAvroSchemaClass.isDefined,
+        "Input is a URI, please specify sample schema via the --avro-schema option.")
+      val sampleClass = Class.forName(sampleAvroSchemaClass.get)
```
nevillelyh: Just double check if I fully understand this. For Avro you don't specify `fields: List[String]`, but instead a trimmed schema containing only fields you want to sample? This will make it hard to use from command line right?
How about stealing the schema projection logic from parquet-avro?

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+                                     output: URI,
+                                     schema: Schema,
+                                     sampleSchema: Class[_])
+
+  // scalastyle:off cyclomatic.complexity
+  private def hashAvroField(r: GenericRecord, fields: Seq[Field], hasher: Hasher): Hasher = {
+    fields.foldLeft(hasher) { case (h, field) =>
+      //TODO: is there need to check for nulls in v?
+      val v = r.get(field.name())
+      field.schema().getType match {
```
ravwojdyla: 👍

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -0,0 +1,346 @@
...
+          throw new UnsupportedOperationException("Null is not supported as sampling key!")
+      }
+    }
+  }
+  // scalastyle:on cyclomatic.complexity
+
+  private def diceElement[T](e: T,
+                             hash: HashCode,
+                             sampleSize: Int,
+                             module: Int): TraversableOnce[T] = {
```
ravwojdyla: 👍

src/main/scala/com/spotify/ratatool/io/FileStorage.scala
```diff
@@ -0,0 +1,186 @@
+/*
```
ravwojdyla: Yep this is mostly scio + some extras.

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
ravwojdyla: Some resources in:
 * http://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed
 * http://stackoverflow.com/questions/11899616/murmurhash-what-is-it

afaiu there won't be guarantee, but it should be reasonably good distribution on different types while still performant. Also the law of big numbers should kick in as well, that said there might be corner cases (highly skewed data) where this will get more data then specified, that said i think that is fine, the `pct` is "approximate" and I believe it's fine that way.

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
ravwojdyla: Yep - correct - this just approximation, which i think is fine.

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
andrewsmartin: Cool thanks for the links!

src/main/scala/com/spotify/ratatool/io/FileStorage.scala
ravwojdyla: `SeekableInput` comes from avro, and this signature is copy pasted from scio.

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
ravwojdyla: 👍

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
ravwojdyla: 👍

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
ravwojdyla: Was thinking of this, but decided against it for now, as it was simpler to keep it in scala for now. Might revisit later if this is a bottleneck. Does it make sense?

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
ravwojdyla: 👍 fixed

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
ravwojdyla: 👍

src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
ravwojdyla: Changed to accept `fields` for avro as well, and dropped the schema idea.

PR #17, opened 2017-08-22
Fix scalacheck `Gen` for Avro `fixed` type

kanterov:


project/plugins.sbt
```diff
@@ -11,3 +11,9 @@ addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "1.1")
 libraryDependencies ++= Seq(
   "com.github.os72" % "protoc-jar" % "3.3.0.1"
 )
+
+// sbt-avro-1-7 uses avro 1.7, so we bump it to 1.8.1
+libraryDependencies ++= Seq(
```
nevillelyh: Should use `version in avroConfig` instead?

kanterov: Hm.. doesn't matter, works in both ways

kanterov: Hm.. but it isn't there, this is inner project that doesn't know anything about outer

kanterov: It work work if `sbt-avro` was in `project/project/plugins.sbt`

kanterov: I guess it work work when `sbt-avro-1-8` will be released

kanterov:
@nevillelyh any chance for release? :)

nevillelyh:
Just released!

PR #19, opened 2017-08-24
Fix scalacheck `Gen` for Avro `map` type

kanterov:


kanterov:
Tests fail because there is something fishy with equality for java hash maps :(

kanterov:
Actually scalacheck generator works property because it uses strings, the problem is in avro `RandomData`, https://github.com/apache/avro/blob/master/lang/java/ipc/src/test/java/org/apache/avro/RandomData.java#L87

PR #20, opened 2017-09-06
Rav/fix assembly and travis

ravwojdyla:


ravwojdyla:
@nevillelyh @kanterov ptal

PR #25, opened 2017-10-23
Upgrade Scala to 2.12.4 and upgrade dependencies

leahxu:
@ravwojdyla

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: scala
 scala:
-  - 2.11.11
+  - 2.12.4
 jdk: oraclejdk8
 
 cache:
@@ -31,5 +31,5 @@ deploy:
   file: "ratatool/target/scala-*/ratatool-*.jar"
   skip_cleanup: true
   on:
-    scala: 2.11.11
+    scala: 2.12.4
     tags: true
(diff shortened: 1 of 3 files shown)


ravwojdyla:
We would need to configure a cross build for both scala 2.11.x and 2.12.x, you could take a look at featran or scio to see an example of cross build setup.

leahxu:
Thanks @ravwojdyla , 2.12 to the crossbuild

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: scala
 scala:
-  - 2.11.11
```
ravwojdyla: Could you specify both `2.11.11` and `2.12.4`?

.travis.yml
```diff
@@ -31,5 +31,5 @@ deploy:
   file: "ratatool/target/scala-*/ratatool-*.jar"
   skip_cleanup: true
   on:
-    scala: 2.11.11
+    scala: 2.12.4
```
ravwojdyla: Let's keep 2.11.11 for the ratatool artifact tool for now.

ravwojdyla:
Thanks @leahxu !

Issue #31, opened 2018-02-02
Upgrade protobuf compiler (protoc) to version 3 and support V3 protobuf messages


annaks:
Currently compiling protobuf protoc with [`-v261`](https://github.com/spotify/ratatool/blob/master/build.sbt#L122) despite depending on the latest [protoc-jar](https://github.com/spotify/ratatool/blob/master/project/plugins.sbt#L12) (`3.3.0.1`).

Across compilers v2 and v3, the internal representation of a generic message was changed from `GeneratedMessage` to `GeneratedMessageV3` which means functions that expect evidence of that type will fail when compiled with v3.  For example, [`protobufOf`](https://github.com/spotify/ratatool/blob/master/ratatool/src/main/scala/com/spotify/ratatool/generators/ProtoBufGenerator.scala#L51) expects `GeneratedMessage` so users of ratatool compiling with v3 will not be able to use this function.

Both [`GeneratedMessage`](https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/GeneratedMessage.java) and [`GeneratedMessageV3`](https://github.com/google/protobuf/blob/master/java/core/src/main/java/com/google/protobuf/GeneratedMessageV3.java) extend `AbstractMessage`.  Suggestion is to support both compiler versions by having evidence of `AbstractMessage`.

PR #32, opened 2018-02-05
use AbstractMessage: super of GeneratedMessage

annaks:
To be compatible with proto 2 and 3 compiled message classes.  Upgrade compiler to v3.3.0

Suggested solution to https://github.com/spotify/ratatool/issues/31.  Maybe it would be better to include implementation for builders and instead change this to `GeneratedMessageV3` for `.asInstanceOf[]` calls?

@idreeskhan


idreeskhan:
I think this looks good. Probably better to maintain resemblance with other Diffy, and the `asInstanceOf` should still invoke subclass method here. 👍

Issue #27, opened 2017-11-06
Make all generators deterministic, refactor package


nevillelyh:
Similar to Avro, make all other `Gen[T]` deterministic.

idreeskhan:
Is the latter half of this title addressed by #35 ?

nevillelyh:
Yep

idreeskhan:
Fixed by #37

PR #37, opened 2018-02-14
Add ratatool-scalacheck generators for TableRow and ProtoBuf

idreeskhan:
Addresses #27, uses Scalacheck `Gen[T]` Monad to create generators.


ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/ProtoBufGenerator.scala
```diff
@@ -0,0 +1,189 @@
...
+      case Type.UINT32 => genPositiveInt.map(v => UInt32Writer(id, v))
+      case Type.UINT64 => genPositiveLong.map(v => UInt64Writer(id, v))
+      case Type.SINT32 => genPositiveInt.map(v => SInt32Writer(id, v))
+      case Type.SINT64 => genPositiveLong.map(v => SInt64Writer(id, v))
+      case Type.FIXED32 => genPositiveInt.map(v => Fixed32Writer(id, v))
+      case Type.FIXED64 => genPositiveLong.map(v => Fixed64Writer(id, v))
+      case Type.SFIXED32 => genPositiveInt.map(v => SFixed32Writer(id, v))
+      case Type.SFIXED64 => genPositiveLong.map(v => SFixed64Writer(id, v))
+      case Type.BOOL => Arbitrary.arbBool.arbitrary.map(v => BoolWriter(id, v))
+      case Type.STRING => Gen.choose(0, 40)
```
jbigred1: what's this magic range for

should we add a comment to explain

jbigred1:
looks good 👍

ankscircle:
:shipit:

ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/ProtoBufGenerator.scala
idreeskhan: Mostly to mimic behaviour here https://github.com/spotify/ratatool/blob/master/ratatool/src/main/scala/com/spotify/ratatool/generators/ProtoBufGenerator.scala#L85

I'm not certain why this is limited to begin with though. @nevillelyh do you know?

idreeskhan: I'll move to an arbitrary string for now, can revert later if needed

PR #49, opened 2018-03-20
Gen Examples

idreeskhan:


ratatool-examples/src/main/scala/com/spotify/ratatool/examples/Examples.scala
```diff
@@ -0,0 +1,134 @@
...
+import org.apache.avro.util.Utf8
+import org.scalacheck.{Arbitrary, Gen}
+import scala.collection.JavaConverters._
+
+object ExampleAvroGen {
+  private val utfGen: Gen[Utf8] = Arbitrary.arbString.arbitrary.map(new Utf8(_))
+
+  private val kvGen: Gen[(Utf8, Int)] = for {
+    k <- utfGen
+    v <- Arbitrary.arbInt.arbitrary
```
jbigred1: can you use `intGen` here instead

ratatool-examples/src/main/scala/com/spotify/ratatool/examples/Examples.scala
```diff
@@ -0,0 +1,134 @@
...
+
+  /**
+   * An example of generating avro data with specific requirements.
+   *
+   * See [[com.spotify.ratatool.avro.specific.ExampleRecord]] for documentation on field
+   * requirements, dependencies, and bounds.
+   */
+  val exampleRecordGen: Gen[ExampleRecord] =
+    specificRecordOf[ExampleRecord]
+      .amend(nestedRecordGen)(_.setNestedRecordField)
```
jbigred1: do you want to make this example less generic make take a use case and generate some stuff for it

ratatool-examples/src/test/scala/com/spotify/ratatool/examples/ExampleAvroGenTest.scala
```diff
@@ -0,0 +1,57 @@
...
+      m.getDependentEnumField == EnumField.Failure) :| "Is Failure on Exception" ||
+    (!m.getIndependentStringField.toString.startsWith("Exception") &&
+      m.getDependentEnumField == EnumField.Success) :| "Is Success when non-Exception"
+  }
+
+  property("double field within bounds") = forAll(gen) { m =>
+    m.getBoundedDoubleField <= 1.0 && m.getBoundedDoubleField >= -1.0
+  }
+
+  property("map field size within bounds") = forAll(gen) { m =>
```
jbigred1: this is interesting

property based testing for a generator

should we recommend people to test their generators when they modify their own code using property based tests

ratatool-examples/src/test/scala/com/spotify/ratatool/examples/ExampleAvroGenTest.scala
```diff
@@ -0,0 +1,57 @@
...
+
+  property("generates valid dependent int") = forAll(gen) { m =>
+    (m.getIndependentIntField == 0
+      && m.getDependentIntField == Int.MaxValue) :| "Max if indep is 0" ||
+    (m.getIndependentIntField != 0
+      && m.getDependentIntField == m.getIndependentIntField/2) :| "Half when indep is not 0"
+  }
+
+  property("generates valid dependent enum") = forAll(gen) { m =>
+    (m.getIndependentStringField.toString.startsWith("Exception") &&
```
jbigred1: can you explain why this should be a string like this

ratatool-examples/src/test/scala/com/spotify/ratatool/examples/ExampleAvroGenTest.scala
idreeskhan: Potentially yes, I think it's dependent on the case and complexity.

ratatool-examples/src/main/scala/com/spotify/ratatool/examples/Examples.scala
idreeskhan: In this case nested record contains a map which has to adhere to certain bounds (i.e. the automatically generated things will not work for it)

PR #52, opened 2018-03-21
BigDiffy output with header

idreeskhan:
```
numTotal	numSame	numDiff	numMissingLhs	numMissingRhs
215273	0	40	9488	205745
```

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -375,7 +375,8 @@ object BigDiffy {
         "key\tdifftype")
 
     result.fieldStats.map(_.toString).saveAsTextFileWithHeader(s"$output/fields",
-        "field\tcount\tfraction\tdeltaStats")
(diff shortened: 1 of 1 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -319,6 +322,20 @@ object BigDiffy {
     (r: TableRow) => get(xs, 0, r)
   }
 
+  def pathWithShards(path: String): String = path.replaceAll("\\/+$", "") + "/part"
+
+  implicit class TextFileHeader(coll: SCollection[String]) {
+    def saveAsTextFileWithHeader(path: String, header: String): Future[Tap[String]] = {
+      val transform = TextIO.write()
+        .to(pathWithShards(path))
+        .withSuffix(".txt")
```
jbigred1: do we want this to be `.tsv` since that's what it produces

idreeskhan: Theoretically it could be, but the existing code outputs `.txt` and I'm moving this header to be optional. I'd rather keep the file format/naming consistent.

Issue #53, opened 2018-03-26
Clearer documentation on usage of BigDiffy, BigSampler


idreeskhan:
Readme should have clear example and usage

Issue #34, opened 2018-02-12
Improve doc and packaging for BigDiffy


nevillelyh:
Right now there's a one liner description of BigDiffy in README. The class is in the release assembly jar but not obvious to users since the mainClass is `Tool` and one has to run `java -cp ratatool.jar ...BigDiffy...`. We should improve this.

PR #65, opened 2018-03-30
Add more documentation on Diffy, Sampler usage

idreeskhan:
#53 #34

### DIFF
--- a/ratatool-diffy/diffy.md
+++ b/ratatool-diffy/diffy.md
@@ -8,9 +8,9 @@ Diffy contains record diff-ing classes that can be utilized by BigDiffy to perfo
 BigDiffy will run a [Scio](https://github.com/spotify/scio) pipeline diffing a LHS and RHS by key
  and output statistics based on the differences. The output contains 3 folders:
 
- * `global` - Global counts of Diff types (`SAME`, `DIFFERENT`, `MISSING_LHS`, `MISSING_RHS`) seen in the entire dataset
- * `fields` - Field level statistics including, but not limited to, the number of records with different values per field, min, max, standard deviation 
- * `keys` - All unique keys found in the two datasets and their Diff types by key in pairs of `<key>\t<diffType>`
+ * `global` - Global counts of Diff types (`SAME`, `DIFFERENT`, `MISSING_LHS`, `MISSING_RHS`) seen in the entire dataset (See `GlobalStats`)
+ * `fields` - Field level statistics including, but not limited to, the number of records with different values per field, min, max, standard deviation (See `FieldStats` and `DeltaStats`)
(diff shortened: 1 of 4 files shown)


ratatool-diffy/diffy.md
```diff
@@ -0,0 +1,31 @@
...
+
+ * `global` - Global counts of Diff types (`SAME`, `DIFFERENT`, `MISSING_LHS`, `MISSING_RHS`) seen in the entire dataset
+ * `fields` - Field level statistics including, but not limited to, the number of records with different values per field, min, max, standard deviation 
+ * `keys` - All unique keys found in the two datasets and their Diff types by key
+
+For full details see [BigDiffy.scala](https://github.com/spotify/ratatool/blob/master/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala)
+
+## Usage
+
+```
```
jbigred1: do we want to discuss the output and how to read it as well

ratatool-sampling/sampler.md
```diff
@@ -0,0 +1,24 @@
+Sampler
+=======
+
+Diffy contains record sampling classes for Avro, BigQuery, Parquet.
+
+# BigSampler
+
+BigSampler will run a [Scio](https://github.com/spotify/scio) pipeline sampling either Avro or BigQuery data.
```
jbigred1: can we note that it supports `gs://` paths as well

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -96,15 +113,17 @@ object BigSampler {
...
-    val output = args("output")
-    val fields = args.list("fields")
-    val seed = args.optional("seed")
+    val (samplePct, input, output, fields, seed) = try {
+      val pct = args("sample").toFloat
+      require(pct > 0.0F && pct <= 1.0F)
+      (pct, args("input"), args("output"), args.list("fields"),
+        args.optional("seed"))
+    } catch {
+      case e: Throwable =>
```
jbigred1: do we want to catch only `NonFatal` here instead of everything

http://www.scala-lang.org/api/current/scala/util/control/NonFatal$.html

ratatool-diffy/diffy.md
idreeskhan: It's discussed a bit in the earlier section. Do you think it needs to be more detailed?

jbigred1: Yeah maybe give an example of how to read the output since I know we mentioned it's hard to parse potentially

ratatool-sampling/sampler.md
idreeskhan: Added

PR #69, opened 2018-04-04
Change Key Stats to output field level deltas and distances

idreeskhan:
#69 Allows user to consume output data for key-level analysis. Also directly puts differences into the output

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/Diffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/Diffy.scala
@@ -42,7 +42,7 @@ case object UnknownDelta extends DeltaValue {
 
 /** Delta value with a known type and computed difference. */
 case class TypedDelta(deltaType: DeltaType.Value, value: Double) extends DeltaValue {
-  override def toString: String = s"$deltaType\t$value}"
+  override def toString: String = s"$deltaType\t$value"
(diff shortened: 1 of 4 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/Diffy.scala
```diff
@@ -36,10 +36,14 @@ object DeltaType extends Enumeration {
...
 /** Delta value of unknown type. */
-case object UnknownDelta extends DeltaValue
+case object UnknownDelta extends DeltaValue {
+  override def toString: String = "UNKNOWN"
+}
 
 /** Delta value with a known type and computed difference. */
-case class TypedDelta(deltaType: DeltaType.Value, value: Double) extends DeltaValue
+case class TypedDelta(deltaType: DeltaType.Value, value: Double) extends DeltaValue {
+  override def toString: String = s"$deltaType\t$value}"
```
jbigred1: did you mean to have a close bracket here on the end of the string

idreeskhan: Heh, no I did not

idreeskhan: Nice catch

Issue #70, opened 2018-04-09
Generator Examples should have a short writeup README


idreeskhan:
This should make it easier to onboard new users by referring to documentation

PR #71, opened 2018-04-12
Add short writeup regarding examples

idreeskhan:
#70


idreeskhan:
Oh I was fiddling with version, will revert

PR #73, opened 2018-04-13
Downgrade guava to 0.20 to maintain consistency with Scio

idreeskhan:


idreeskhan:
@jbigred1 @tkhduracell

tkhduracell:
👍

Issue #74, opened 2018-04-13
Look into reducing Guava deps


idreeskhan:
Mostly it is pulled it to maintain consistency across libs but this potentially causes issues. Should investigate further

idreeskhan:
Closing as duplicate of #72 since changes for that will already address this

Issue #72, opened 2018-04-12
Keep guava in step with Scio


idreeskhan:
Currently we are on 0.21 while Scio is on 0.20. Should investigate downgrade and potential issues. Concerned this may cause unexpected errors when running.

idreeskhan:
Seems to be the case. Fixing now

idreeskhan:
Closing, please try 0.3.1 and re-open if there are issues

tkhduracell:
@idreeskhan Still reproducible with the 0.3.1.jar 
```
[ForkJoinPool-1-worker-5] ERROR org.apache.beam.runners.dataflow.util.MonitoringUtil$LoggingHandler - 2018-04-13T15:11:52.057Z: java.lang.NoSuchMethodError: com.google.common.base.Preconditions.checkNotNull(Ljava/lang/Object;Ljava/lang/String;Ljava/lang/Object;)Ljava/lang/Object;
	at org.apache.beam.sdk.io.gcp.bigquery.BigQueryAvroUtils.convertRequiredField(BigQueryAvroUtils.java:169)
	at org.apache.beam.sdk.io.gcp.bigquery.BigQueryAvroUtils.getTypedCellValue(BigQueryAvroUtils.java:131)
	at org.apache.beam.sdk.io.gcp.bigquery.BigQueryAvroUtils.convertGenericRecordToTableRow(BigQueryAvroUtils.java:115)
	at org.apache.beam.sdk.io.gcp.bigquery.BigQueryAvroUtils.convertGenericRecordToTableRow(BigQueryAvroUtils.java:104)
	at org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO$TableRowParser.apply(BigQueryIO.java:381)
	at org.apache.beam.sdk.io.gcp.bigquery.BigQueryIO$TableRowParser.apply(BigQueryIO.java:374)
	at org.apache.beam.sdk.io.gcp.bigquery.BigQuerySourceBase$1.apply(BigQuerySourceBase.java:204)
	at org.apache.beam.sdk.io.gcp.bigquery.BigQuerySourceBase$1.apply(BigQuerySourceBase.java:198)
	at org.apache.beam.sdk.io.AvroSource$AvroBlock.readNextRecord(AvroSource.java:581)
	at org.apache.beam.sdk.io.BlockBasedSource$BlockBasedReader.readNextRecord(BlockBasedSource.java:223)
[...]

PR #75, opened 2018-04-13
Remove Guava dep entirely

idreeskhan:
#72 #74 @tkhduracell


tkhduracell:
Line `val guavaVersion = "20.0"`can also be removed

idreeskhan:
👍 Will remove in separate commit, just want to have a working JAR out

Issue #77, opened 2018-04-16
Remove assembly steps from build.sbt


idreeskhan:
Also involves updating Travis. Removing in favour of `sbt pack` as a way to simplify dep issues and more cleanly address #72

Issue #76, opened 2018-04-13
BigSampler fails to run with fat jar


idreeskhan:
```
Exception in thread "main" java.lang.IllegalAccessError: tried to access class org.apache.beam.sdk.io.gcp.bigquery.BigQueryServicesImpl from class com.spotify.ratatool.samplers.BigSamplerBigQuery$
	at com.spotify.ratatool.samplers.BigSamplerBigQuery$.sampleBigQueryTable(BigSampler.scala:427)
	at com.spotify.ratatool.samplers.BigSampler$.singleInput(BigSampler.scala:142)
	at com.spotify.ratatool.samplers.BigSampler$.main(BigSampler.scala:162)
	at com.spotify.ratatool.samplers.BigSampler.main(BigSampler.scala)
```

idreeskhan:
#72 fix will likely be moving to `sbt pack` based solution, which should solve this problem as well.

idreeskhan:
Moving to `sbt pack`, currently there is `0.3.2-SNAPSHOT` available for this and will be releasing in this manner in the future

Issue #67, opened 2018-04-02
Tool/CLI should provide a cleaner wrapper around other parts of the project


idreeskhan:
Currently it is mainly a wrapper for sampler, but the jar can be directly invoked to run `BigDiffy` or `BigSampler`. It would be better to have the CLI wrap all of this functionality to be easier to use.

idreeskhan:
README should also be updated to reflect the new method of running since we will move to `sbt pack`

PR #78, opened 2018-04-16
Move to SBT Pack instead of Assembly

idreeskhan:
#78 
#77 
#76 
and potentially #67 

We want to be able to move to SBT Pack instead of assembly to simplify the dependency management and avoid conflicts with building the fat jar. To make the usage easier, we reduce to a single main so that there is only one entry point, and those who brew install can still use `ratatool [args]` to run commands

Should keep https://github.com/spotify/homebrew-public/pull/24 in sync and merge that PR when this one is merged


build.sbt
```diff
@@ -174,7 +160,6 @@ lazy val ratatoolCli = project
     name := "ratatool-cli",
     libraryDependencies ++= Seq(
       "com.github.scopt" %% "scopt" % scoptVersion,
-      "org.apache.parquet" % "parquet-avro" % parquetVersion,
```
jbigred1: did we mean to remove this

ratatool-cli/src/main/scala/com/spotify/ratatool/tool/Ratatool.scala
```diff
@@ -0,0 +1,74 @@
...
+object Ratatool {
+
+  //scalastyle:off cyclomatic.complexity
+  def main(args: Array[String]): Unit = {
+    val usage = """
+      | Ratatool - a tool for random data generation and sampling
+      | Usage: ratatool [bigDiffy|bigSampler|directSampler] [args]
+    """.stripMargin
+
+    if (args.isEmpty || !Set("bigDiffy", "bigSampler", "directSampler").contains(args.head)) {
```
jbigred1: can we make these variables constants

ratatool-cli/src/main/scala/com/spotify/ratatool/tool/Ratatool.scala
```diff
@@ -0,0 +1,74 @@
...
+import com.spotify.ratatool.io.{AvroIO, BigQueryIO, ParquetIO, TableRowJsonIO}
+import com.spotify.ratatool.samplers.{AvroSampler, BigQuerySampler, BigSampler, ParquetSampler}
+import org.apache.hadoop.fs.Path
+
+object Ratatool {
+
+  //scalastyle:off cyclomatic.complexity
+  def main(args: Array[String]): Unit = {
+    val usage = """
+      | Ratatool - a tool for random data generation and sampling
```
jbigred1: what about diffing

ratatool-cli/src/main/scala/com/spotify/ratatool/tool/Ratatool.scala
```diff
@@ -0,0 +1,74 @@
...
+        case "bigDiffy" => BigDiffy.run(args.tail)
+        case "bigSampler" => BigSampler.run(args.tail)
+        case "directSampler" =>
+          val opts = DirectSamplerParser.parse(args.tail)
+          if (opts.isEmpty) {
+            sys.exit(-1)
+          }
+
+          val o = opts.get
+          o.mode match {
```
jbigred1: instead of `get` could we do `opts.foreach`

idreeskhan:
Yes definitely need to update docs, will roll that into this PR as well

build.sbt
idreeskhan: Yea I think it ended up there as an artifact during the project restructure but seems to work fine without

ratatool-cli/src/main/scala/com/spotify/ratatool/tool/Ratatool.scala
idreeskhan: Nice catch

ratatool-cli/src/main/scala/com/spotify/ratatool/tool/Ratatool.scala
idreeskhan: 👍

Issue #82, opened 2018-04-23
Update Scalacheck to 1.14.0


idreeskhan:
Includes initial seed setting, which can potentially enable #81

PR #84, opened 2018-04-23
Scio, Beam, Scalacheck Version bumps

idreeskhan:
#56 #82 

I suspect we will hit Travis issues due to DirectRunner slowness, so will include fixing that in this PR as well


ratatool-sampling/src/test/scala/com/spotify/ratatool/samplers/BigSamplerTest.scala
```diff
@@ -211,11 +211,11 @@ object BigSamplerTest extends Properties("BigSampler") {
...
 
   val schema = Schemas.avroSchema
-  val data1 = Gen.listOfN(40000, genericRecordOf(schema))
+  val data1 = Gen.listOfN(20000, genericRecordOf(schema))
     .pureApply(Gen.Parameters.default.withSize(5), Seed.random())
-  val data2 = Gen.listOfN(10000, genericRecordOf(schema))
+  val data2 = Gen.listOfN(5000, genericRecordOf(schema))
     .pureApply(Gen.Parameters.default.withSize(5), Seed.random())
-  val totalElements = 50000
+  val totalElements = 25000
```
jbigred1: do you want to add a comment as to what a reasonable maximum we can use here is for Travis issues

idreeskhan: I think it's too finicky and likely to change frequently with underlying implementation. As long the test reasonably tests functionality I don't care too much about what this number is.

Issue #24, opened 2017-09-26
Use farmhash instead of murmur for BigSampler


ravwojdyla:
https://github.com/google/farmhash is available in guava, this would allow us to use build in hashing in BigQuery (https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-and-operators#farm_fingerprint).

anne-decusatis:
https://google.github.io/guava/releases/23.0/api/docs/com/google/common/hash/Hashing.html#farmHashFingerprint64-- says: "FarmHashFingerprints generated using this are byte-wise identical to those created using the C++ version, but note that this uses unsigned integers (see UnsignedInts). Comparisons between the two should take this into account."

The BigQuery examples which use FarmHash have both negative and positive numbers returned (though of course I do not know whether it uses C++ or Java or something else entirely under the hood). 

I will investigate more to make sure that the BigQuery farmhash fingerprints are actually comparable to the Guava farmhash fingerprints before implementing.

anne-decusatis:
in Guava: 
```
scala> Hashing.farmHashFingerprint64().newHasher().putString("test", Charset.forName("utf-8")).hash.asLong
res14: Long = 8581389452482819506
```

in BigQuery:
```select farm_fingerprint("test");```
```

Row | f0_ |  
-- | -- | --
1 | 8581389452482819506 |  
```

These look the same to me!

idreeskhan:
If they are bytewise identical it should not matter much. Have you found a case where BQ returns a negative? Would be interesting to see what is produced by a mod operation on that since that is how we would be using it.

idreeskhan:
Best to proceed as is since we want it to produce the same output as BQ

PR #92, opened 2018-05-14
replace murmur hash with farmhash

anne-decusatis:
Addresses #24 

Not sure if more communication around this potentially being a breaking change is necessary

### DIFF
--- a/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
+++ b/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
@@ -53,11 +53,8 @@ object BigSampler extends Command {
   private[samplers] val utf8Charset = Charset.forName("UTF-8")
   private[samplers] val fieldSep = '.'
 
-  private[samplers] def kissHashFun(seed: Option[Int] = None): HashFunction = {
-    seed match {
-      case Some(s) => Hashing.murmur3_128(s)
-      case None => Hashing.murmur3_128(System.currentTimeMillis.toInt)
-    }
+  private[samplers] def kissHashFun(): HashFunction = {
+    Hashing.farmHashFingerprint64() // doesn't require seed, returns same fingerprint each time
   }
 
   /**
@@ -92,7 +89,6 @@ object BigSampler extends Command {
         |  --input=<path>                    Input file path or BigQuery table
         |  --output=<path>                   Output file path or BigQuery table
         |  [--fields=<field1,field2,...>]    An optional list of fields to include in hashing for sampling cohort selection
-        |  [--seed=<seed]                    An optional seed using in hashing for sampling cohort selection
       """.stripMargin)
     // scalastyle:on regex line.size.limit
     sys.exit(1)
@@ -115,11 +111,10 @@ object BigSampler extends Command {
     val (sc, args) = ContextAndArgs(argv)
     val (opts, _) = ScioContext.parseArguments[PipelineOptions](argv)
 
-    val (samplePct, input, output, fields, seed) = try {
+    val (samplePct, input, output, fields) = try {
       val pct = args("sample").toFloat
       require(pct > 0.0F && pct <= 1.0F)
-      (pct, args("input"), args("output"), args.list("fields"),
-        args.optional("seed"))
+      (pct, args("input"), args("output"), args.list("fields"))
     } catch {
       case e: Throwable =>
         usage()
@@ -128,10 +123,6 @@ object BigSampler extends Command {
 
     if (fields.isEmpty) {
       log.warn("No fields to hash on specified, won't guarantee cohorts between datasets.")
-    }
-
-    if (seed.isEmpty) {
-      log.warn("No seed specified, won't guarantee cohorts between datasets.")
     }
 
     if (parseAsBigQueryTable(input).isDefined) {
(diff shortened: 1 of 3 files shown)


ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -53,11 +53,8 @@ object BigSampler extends Command {
   private[samplers] val utf8Charset = Charset.forName("UTF-8")
   private[samplers] val fieldSep = '.'
 
-  private[samplers] def kissHashFun(seed: Option[Int] = None): HashFunction = {
-    seed match {
-      case Some(s) => Hashing.murmur3_128(s)
-      case None => Hashing.murmur3_128(System.currentTimeMillis.toInt)
-    }
+  private[samplers] def kissHashFun(): HashFunction = {
```
idreeskhan: Should be renamed to `farmHashFun`

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -92,7 +89,6 @@ object BigSampler extends Command {
         |  --input=<path>                    Input file path or BigQuery table
         |  --output=<path>                   Output file path or BigQuery table
         |  [--fields=<field1,field2,...>]    An optional list of fields to include in hashing for sampling cohort selection
-        |  [--seed=<seed]                    An optional seed using in hashing for sampling cohort selection
```
idreeskhan: Thinking a bit more about our discussion earlier I am backtracking on what I said, I think it's actually best to leave this and the original hash function in for backwards compatibility. We can default to Farm hasher and allow specifying a param to use the original hasher. Only question is how to handle the seed with farm hash, potentially it can be appended to the fields we are hashing over.

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
idreeskhan: Same goes for all the references elsewhere.

ratatool-sampling/src/test/scala/com/spotify/ratatool/samplers/BigSamplerTest.scala
```diff
@@ -45,11 +45,21 @@ object BigSamplerTest extends Properties("BigSampler") {
     val dice2 = BigSampler.diceElement(i, hasher2.hash(), 1.0)
     dice1 == dice2
   }
+`
+  private val negativeHashCodeStringGen: Gen[String] = Gen.alphaStr
+    .suchThat(str => { newTestHasher().putString(str, BigSampler.utf8Charset).hash().asLong < 0} )
+  property("dice on hash values that return negative Long representations should still bucket OK") =
```
idreeskhan: This is awesome, can we have a more specific title? Or just a better word that `OK`

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
anne-decusatis: I think we should ignore it rather than use it as salt (as you're describing) -- unless you think it's important to be able to have different output with different seeds? I'm not as sure of the use case as you are...

idreeskhan: Yea it allows for it. If no seed is specified it will be vanilla farm hash. We can change later if it's confusing, but I think this is more clear than a param which does nothing or having the CLI have nested params

anne-decusatis: I addressed your other comments but I think I misunderstood what you meant by this one. Let me know if you want to talk more about it!

idreeskhan:
Also tests are failing on scalastyle https://travis-ci.org/spotify/ratatool/jobs/378916482#L565

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
anne-decusatis: will do, thanks!

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -454,7 +471,7 @@ private[samplers] object BigSamplerBigQuery {
         val samplePct100 = samplePct * 100.0
         val r = sc.bigQueryTable(inputTbl)
           .flatMap { e =>
-            val hasher = BigSampler.kissHashFun(seed).newHasher(fields.size)
```
idreeskhan: What is the effect of removing the `newHasher`? As I understand it shouldn't matter much but should verify.

ratatool-sampling/src/test/scala/com/spotify/ratatool/samplers/BigSamplerTest.scala
```diff
@@ -114,7 +136,7 @@ object BigSamplerTest extends Properties("BigSampler") {
     "double_field")
 
   property("should hash on supported required types in Avro") = forAll(richAvroGen) { r =>
-    val hasher = newTestHasher()
+    val hasher = newTestFarmHasher()
```
idreeskhan: Not necessarily in the scope of this PR but would be nice to either map across both hashers or gen them instead.

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
anne-decusatis: I changed it to pass around the `Hasher` (essentially a builder for hashes) once it was already generated, rather than the `HashFunction` which initializes a hasher. This was necessary because we want to salt FarmHash with the "seed" param once it's already created, but we only know about the seed when we're initializing, not when we're adding existing fields (and we don't want to care about the "seed" as a murmur seed vs as farmhash salt when we're adding fields, because that would leak hash implementation details)

ratatool-sampling/src/test/scala/com/spotify/ratatool/samplers/BigSamplerTest.scala
anne-decusatis: I"m not sure it's in scope either but it sounds fun and relatively quick to use Gen.oneOf so I'll see what I can do

Issue #94, opened 2018-05-18
BigDiffy should allow for nested unordered repeated fields


idreeskhan:
Currently if a field is `REPEATED` and specified in `unordered`, the list is sorted by doing `.toString`. This works for some cases but breaks down when the there is an inner field which is repeated and should be unordered.

Example:
```
first_repeated: [
  {key: firstval, second_repeated:[a, b, c]},
  {key: secondval, second_repeated:[1,2,3]}
]
```
diffed with
```
first_repeated: [
  {key: secondval, second_repeated:[2,1,3]},
  {key: firstval, second_repeated:[a, c, b]}
]
```

In this cause assuming `unordered = Set("first_repeated", "first_repeated.second_repeated")`. Currently on `first_repeated` it would convert `.toString` and then produce a Delta, even though a user would expect them to resolve to the same thing.

PR #99, opened 2018-05-31
Deeply Nested Repeated Diffing

idreeskhan:
#94 
To do: Docs


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
```diff
@@ -34,6 +35,10 @@ class AvroDiffy[T <: GenericRecord](ignore: Set[String] = Set.empty,
 
   // scalastyle:off cyclomatic.complexity
   private def diff(x: GenericRecord, y: GenericRecord, root: String): Seq[Delta] = {
+    def getField(f: String)(x: GenericRecord): AnyRef = {
+      x.get(f)
+    }
```
jbigred1: can probably make this a one line expression

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
```diff
@@ -49,9 +54,22 @@ class AvroDiffy[T <: GenericRecord](ignore: Set[String] = Set.empty,
             diff(a, b, fullName)
           }
         case Schema.Type.ARRAY if unordered.contains(fullName) =>
-          val a = sortList(x.get(name).asInstanceOf[java.util.List[AnyRef]])
-          val b = sortList(y.get(name).asInstanceOf[java.util.List[AnyRef]])
-          if (a == b) Nil else Seq(Delta(fullName, a, b, delta(a, b)))
+          if (f.schema().getElementType.getType == Schema.Type.RECORD
+              && unordered.exists(_.startsWith(s"$fullName."))
```
jbigred1: don't we already check a stricter check above `case Schema.Type.ARRAY if unordered.contains(fullName)` is stricter than `unordered.exists(_.startsWith(s"$fullName."))` so this seems unnecessary then potentially

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/ProtoBufDiffy.scala
```diff
@@ -37,28 +38,45 @@ class ProtoBufDiffy[T <: AbstractMessage : ClassTag](ignore: Set[String] = Set.e
...
+
     fields.flatMap { f =>
       val name = f.getName
       val fullName = if (root.isEmpty) name else root + "." + name
       if (f.isRepeated && unordered.contains(fullName)) {
-        val a = sortList(x.getField(f).asInstanceOf[java.util.List[AnyRef]])
-        val b = sortList(y.getField(f).asInstanceOf[java.util.List[AnyRef]])
-        if (a == b) Nil else Seq(Delta(fullName, a, b, delta(a, b)))
+        val a = sortList(x.getField(f).asInstanceOf[java.util.List[AbstractMessage]],
+          unorderedFieldKeys.get(fullName).map(s => getField(getFieldDescriptor(f, s)) _))
```
jbigred1: maybe make this a `val` since we use it a few times

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/ProtoBufDiffy.scala
```diff
@@ -37,28 +38,45 @@ class ProtoBufDiffy[T <: AbstractMessage : ClassTag](ignore: Set[String] = Set.e
...
+        val b = sortList(y.getField(f).asInstanceOf[java.util.List[AbstractMessage]],
+          unorderedFieldKeys.get(fullName).map(s => getField(getFieldDescriptor(f, s)) _))
+        if (f.getJavaType == JavaType.MESSAGE && unordered.exists(_.startsWith(s"$fullName."))
+            && unorderedFieldKeys.contains(fullName)) {
+          a.asInstanceOf[java.util.List[AbstractMessage]].asScala.zip(
+            b.asInstanceOf[java.util.List[AbstractMessage]].asScala).flatMap {
```
jbigred1: could we add comments to this

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/ProtoBufDiffy.scala
```diff
@@ -37,28 +38,45 @@ class ProtoBufDiffy[T <: AbstractMessage : ClassTag](ignore: Set[String] = Set.e
...
-        val a = sortList(x.getField(f).asInstanceOf[java.util.List[AnyRef]])
-        val b = sortList(y.getField(f).asInstanceOf[java.util.List[AnyRef]])
-        if (a == b) Nil else Seq(Delta(fullName, a, b, delta(a, b)))
+        val a = sortList(x.getField(f).asInstanceOf[java.util.List[AbstractMessage]],
+          unorderedFieldKeys.get(fullName).map(s => getField(getFieldDescriptor(f, s)) _))
+        val b = sortList(y.getField(f).asInstanceOf[java.util.List[AbstractMessage]],
+          unorderedFieldKeys.get(fullName).map(s => getField(getFieldDescriptor(f, s)) _))
+        if (f.getJavaType == JavaType.MESSAGE && unordered.exists(_.startsWith(s"$fullName."))
+            && unorderedFieldKeys.contains(fullName)) {
+          a.asInstanceOf[java.util.List[AbstractMessage]].asScala.zip(
```
jbigred1: do we need to keep converting from Java to Scala and back

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
idreeskhan: This is to check if there is a nested field that should be unordered, which is the case where we need to properly sort and recurse to compare. If there is nothing with `fullName.` then the assuming is we can sort this regularly since all fields should be precise

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/ProtoBufDiffy.scala
idreeskhan: Yea in this case it's a Java List that we get from protobuf and we need to cast it before wrapping in Scala collection

Issue #100, opened 2018-06-06
Take in multiple functions in amend


jbigred1:
An example of this could be

```scala
.amend(intGen)(m => i => {
        m.setField1(i)
        m.setField2(dependentIntFunc(i))
      })
```

would become

```scala
.amend(intGen)(,_.setField1, _.setField2)
```

PR #102, opened 2018-06-07
adding in multiple functions

jbigred1:
This will allow it so multiple functions can be provided to the `amend` functions now

Addresses https://github.com/spotify/ratatool/issues/100


ratatool-examples/src/main/scala/com/spotify/ratatool/examples/scalacheck/GenExample.scala
```diff
@@ -96,14 +96,10 @@ object ExampleAvroGen {
...
        */
-      .amend(intGen)(m => i => {
-        m.setIndependentIntField(i)
-        m.setDependentIntField(dependentIntFunc(i))
-      })
-      .amend(stringGen)(m => s => {
-        m.setIndependentStringField(s)
-        m.setDependentEnumField(dependentEnumFunc(s))
-      })
+      .amend(intGen)(_.setIndependentIntField,
```
idreeskhan: Bit of a nitpick but can we add the first function on a newline here, makes it more clear when reading the examples that it's doing two things

PR #113, opened 2018-07-09
Upload diff results from BigDiffy to BigQuery

### LINKED ISSUE #106: Support BigQuery upload of BigDiffy results
The bigdiffy tool is great but isn't very user-friendly. It's great that is can do diffs on BigQuery tables, but it is a bit cumbersome that the results are written to GCS, also without any schema. 

I suggest that we change the CLI API so that one can choose the write results to either GCS or BQ. 
```
big-diffy --runner=DataflowRunner <stuff> --key=record_id --lhs=bq://<dataset1> --rhs=bq://<dataset2> --output=bq://project/dataset/table
```

Also it would be nice to auto infer mode=bigquery in this case. When lhs and rhs has bq:// prefix.

anne-decusatis:
fixes #106

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -25,6 +25,7 @@ import com.spotify.ratatool.{Command, GcsConfiguration}
 import com.spotify.ratatool.samplers.AvroSampler
 import com.spotify.scio._
 import com.spotify.scio.bigquery.BigQueryClient
+import com.spotify.scio.bigquery.types.BigQueryType
 import com.spotify.scio.io.Tap
 import com.spotify.scio.values.SCollection
 import com.twitter.algebird._
@@ -282,19 +283,58 @@ object BigDiffy extends Command {
     r.setFields(mergeFields(x.getFields.asScala, y.getFields.asScala).asJava)
   }
 
-  def saveStats[T](bigDiffy: BigDiffy[T], output: String, withHeader: Boolean = false): Unit = {
-    if (withHeader) {
-      bigDiffy.keyStats.map(_.toString).saveAsTextFileWithHeader(s"$output/keys", "key\tdifftype")
-      bigDiffy.fieldStats.map(_.toString).saveAsTextFileWithHeader(s"$output/fields",
-        "field\tcount\tfraction\tdeltaType\tmin" +
-          "\tmax\tcount\tmean\tvariance\tstddev\tskewness\tkurtosis")
-      bigDiffy.globalStats.map(_.toString).saveAsTextFileWithHeader(s"$output/global",
-        "numTotal\tnumSame\tnumDiff\tnumMissingLhs\tnumMissingRhs")
-    }
-    else {
-      bigDiffy.keyStats.saveAsTextFile(s"$output/keys")
-      bigDiffy.fieldStats.saveAsTextFile(s"$output/fields")
-      bigDiffy.globalStats.saveAsTextFile(s"$output/global")
+  @BigQueryType.toTable
+  case class KeyStatsBigQuery(key: String, diffType: String, deltaString: Option[String])
+  @BigQueryType.toTable
+  case class GlobalStatsBigQuery(numTotal: Long, numSame: Long, numDiff: Long,
+                                 numMissingLhs: Long, numMissingRhs: Long)
+  @BigQueryType.toTable
+  case class FieldStatsBigQuery(field: String,
+                        count: Long,
+                        fraction: Double,
+                        deltaStats: Option[DeltaStatsBigQuery])
+  case class DeltaStatsBigQuery(deltaType: String,
+                        min: Double, max: Double, count: Long,
+                        mean: Double, variance: Double, stddev: Double,
+                        skewness: Double, kurtosis: Double)
+
+  /** saves stats to either GCS as text, or BigQuery */
+  def saveStats[T](bigDiffy: BigDiffy[T], output: String, withHeader: Boolean = false,
+                   saveToBq: Boolean = false): Unit = {
+    val keyStatsPath = s"$output/keys"
+    val fieldStatsPath = s"$output/fields"
+    val globalStatsPath = s"$output/global"
+    if (!saveToBq) {
+      // saving to GCS, either with or without header
+      if (withHeader) {
+        bigDiffy.keyStats.map(_.toString).saveAsTextFileWithHeader(keyStatsPath, "key\tdifftype")
+        bigDiffy.fieldStats.map(_.toString).saveAsTextFileWithHeader(fieldStatsPath,
+          "field\tcount\tfraction\tdeltaType\tmin" +
+            "\tmax\tcount\tmean\tvariance\tstddev\tskewness\tkurtosis")
+        bigDiffy.globalStats.map(_.toString).saveAsTextFileWithHeader(globalStatsPath,
+          "numTotal\tnumSame\tnumDiff\tnumMissingLhs\tnumMissingRhs")
+      }
+      else {
+        bigDiffy.keyStats.saveAsTextFile(keyStatsPath)
+        bigDiffy.fieldStats.saveAsTextFile(fieldStatsPath)
+        bigDiffy.globalStats.saveAsTextFile(globalStatsPath)
+      }
+    } else {
+      // saving to BQ, header irrelevant
+
+      bigDiffy.keyStats.map(stat =>
+        KeyStatsBigQuery(stat.key, stat.diffType.toString, stat.delta.map(_.toString)))
+        .saveAsTypedBigQuery(s"${output}_keys")
+      bigDiffy.fieldStats.map(stat =>
+        FieldStatsBigQuery(stat.field, stat.count, stat.fraction, stat.deltaStats.map(ds =>
+          DeltaStatsBigQuery(ds.deltaType.toString, ds.min, ds.max, ds.count, ds.mean,
+            ds.variance, ds.stddev, ds.skewness, ds.kurtosis)
+        )))
+        .saveAsTypedBigQuery(s"${output}_fields")
+      bigDiffy.globalStats.map( stat =>
+        GlobalStatsBigQuery(stat.numTotal, stat.numSame, stat.numDiff,
+          stat.numMissingLhs, stat.numMissingRhs))
+        .saveAsTypedBigQuery(s"${output}_global")
     }
   }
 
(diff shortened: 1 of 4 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -282,19 +283,68 @@ object BigDiffy extends Command {
...
-        "field\tcount\tfraction\tdeltaType\tmin" +
-          "\tmax\tcount\tmean\tvariance\tstddev\tskewness\tkurtosis")
-      bigDiffy.globalStats.map(_.toString).saveAsTextFileWithHeader(s"$output/global",
-        "numTotal\tnumSame\tnumDiff\tnumMissingLhs\tnumMissingRhs")
-    }
-    else {
-      bigDiffy.keyStats.saveAsTextFile(s"$output/keys")
-      bigDiffy.fieldStats.saveAsTextFile(s"$output/fields")
-      bigDiffy.globalStats.saveAsTextFile(s"$output/global")
+  @BigQueryType.toTable
```
idreeskhan: If paradise import fixes the annotation compilation then we can probably reduce these down to a single class again

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -380,23 +431,32 @@ object BigDiffy extends Command {
...
+          args.list("unordered").toSet, args("output-mode"))
       } catch {
         case e: Throwable =>
           usage()
           throw e
       }
     }
 
-    val result = mode match {
+    val saveToBq: Boolean = outputMode match {
```
idreeskhan: I think I prefer to leave this as a str for 2 reasons. 1 to match the current avro/bq impl for input. 2 so that it's easier to see the effects later on the code. Can use a `case match` instead of `if` later to split it out

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -282,19 +283,68 @@ object BigDiffy extends Command {
...
-          "\tmax\tcount\tmean\tvariance\tstddev\tskewness\tkurtosis")
-      bigDiffy.globalStats.map(_.toString).saveAsTextFileWithHeader(s"$output/global",
-        "numTotal\tnumSame\tnumDiff\tnumMissingLhs\tnumMissingRhs")
-    }
-    else {
-      bigDiffy.keyStats.saveAsTextFile(s"$output/keys")
-      bigDiffy.fieldStats.saveAsTextFile(s"$output/fields")
-      bigDiffy.globalStats.saveAsTextFile(s"$output/global")
+  @BigQueryType.toTable
+  case class KeyStatsBigQuery(key: String, diffType: String, delta: Option[DeltaBigQuery])
```
idreeskhan: Outside the scope of this PR but we can potentially add `@description` annotations here now that it can write out to BQ to make it clearer for users what fields are.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
anne-decusatis: I was also seeing other errors around the subtypes not being mappable to BigQuery types - in particular, I'm saving enums and Any types as Strings in BigQuery with these new case classes. I would prefer to leave them separate.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
anne-decusatis: I can pass it in as a string and parse it in saveStats -- this has the downside that it will still run the analysis & only fail at the very end when saving. The code I have now throws an exception before doing any work if the output mode is invalid, since this branch is early in the main function. What do you think is better? Could also do a validation before running and still pass around the string but I find that a bit harder to follow, personally.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
anne-decusatis: should I create a follow up issue for this?

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -282,19 +283,68 @@ object BigDiffy extends Command {
...
+                        mean: Double, variance: Double, stddev: Double,
+                        skewness: Double, kurtosis: Double)
+
+  /** saves stats to either GCS as text, or BigQuery */
+  def saveStats[T](bigDiffy: BigDiffy[T], output: String, withHeader: Boolean = false,
+                   saveToBq: Boolean = false): Unit = {
+    val keyStatsPath = s"$output/keys"
+    val fieldStatsPath = s"$output/fields"
+    val globalStatsPath = s"$output/global"
+    if (!saveToBq) {
```
tkhduracell: Unnecessary negation?

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -282,19 +283,68 @@ object BigDiffy extends Command {
...
+      // saving to GCS, either with or without header
+      if (withHeader) {
+        bigDiffy.keyStats.map(_.toString).saveAsTextFileWithHeader(keyStatsPath, "key\tdifftype")
+        bigDiffy.fieldStats.map(_.toString).saveAsTextFileWithHeader(fieldStatsPath,
+          "field\tcount\tfraction\tdeltaType\tmin" +
```
tkhduracell: Could maybe we nice with `"\t".join("field", "count", ...)`?

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
tkhduracell: Enum value?

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -282,19 +283,68 @@ object BigDiffy extends Command {
...
+            val typedDelta = d.delta.asInstanceOf[TypedDelta]
+            DeltaValueBigQuery(typedDelta.deltaType.toString, Option(typedDelta.value))
+          } else {
+            DeltaValueBigQuery("UNKNOWN", None)
+          }
+          DeltaBigQuery(d.field, d.left.toString, d.right.toString, dv)})
+        ))
+        .saveAsTypedBigQuery(s"${output}_keys")
+      bigDiffy.fieldStats.map(stat =>
+        FieldStatsBigQuery(stat.field, stat.count, stat.fraction, stat.deltaStats.map(ds =>
```
tkhduracell: Maybe this transformation could live in a companion to FieldStatsBigquery,  or similar approach?

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -282,19 +283,68 @@ object BigDiffy extends Command {
...
+      // saving to GCS, either with or without header
+      if (withHeader) {
+        bigDiffy.keyStats.map(_.toString).saveAsTextFileWithHeader(keyStatsPath, "key\tdifftype")
+        bigDiffy.fieldStats.map(_.toString).saveAsTextFileWithHeader(fieldStatsPath,
+          "field\tcount\tfraction\tdeltaType\tmin" +
+            "\tmax\tcount\tmean\tvariance\tstddev\tskewness\tkurtosis")
+        bigDiffy.globalStats.map(_.toString).saveAsTextFileWithHeader(globalStatsPath,
+          "numTotal\tnumSame\tnumDiff\tnumMissingLhs\tnumMissingRhs")
+      }
+      else {
```
tkhduracell: Should newline be here?

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
idreeskhan: #119

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -282,19 +287,68 @@ object BigDiffy extends Command {
...
+                        deltaStats: Option[DeltaStatsBigQuery])
+  case class DeltaStatsBigQuery(deltaType: String,
+                        min: Double, max: Double, count: Long,
+                        mean: Double, variance: Double, stddev: Double,
+                        skewness: Double, kurtosis: Double)
+
+  /** saves stats to either GCS as text, or BigQuery */
+  def saveStats[T](bigDiffy: BigDiffy[T], output: String, withHeader: Boolean = false,
+                   outputMode: OutputMode = GCS): Unit = {
+    val keyStatsPath = s"$output/keys"
```
jbigred1: can we move these to exist inside the `GCS` match statement

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -329,7 +383,8 @@ object BigDiffy extends Command {
       s"""BigDiffy - pair-wise field-level statistical diff
         |Usage: ratatool $command [dataflow_options] [options]
         |
-        |  --mode=[avro|bigquery]
```
jbigred1: do we want to support `mode` as well for backwards compatibility

idreeskhan: I think it's fine to break it, will add a more helpful log error if `mode` is passed in

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -329,14 +384,15 @@ object BigDiffy extends Command {
...
-        |  --key=<key>            '.' separated key field
-        |  --lhs=<path>           LHS File path or BigQuery table
-        |  --rhs=<path>           RHS File path or BigQuery table
-        |  --output=<output>      File path prefix for output
-        |  --ignore=<keys>        ',' separated field list to ignore
-        |  --unordered=<keys>     ',' separated field list to treat as unordered
-        |  [--with-header]        Output all TSVs with header rows
+        |  --input-mode=(avro|bigquery)     Diff-ing Avro or BQ records
+        |  [--output-mode=(gcs|bigquery)]   Saves to a text file in GCS or a BigQuery dataset
```
catherinejelder: what do the braces connote - optional args?

idreeskhan: Yea, common convention in bash help/usage guide

PR #120, opened 2018-08-15
Bump Scio to 0.5.7

ClaireMcGinty:


build.sbt
```diff
@@ -27,7 +27,7 @@ val parquetVersion = "1.9.0"
 val protoBufVersion = "3.3.1"
 val scalaCheckVersion = "1.14.0"
 val scalaTestVersion = "3.0.4"
-val scioVersion = "0.5.6"
+val scioVersion = "0.5.7"
```
idreeskhan: Will also need to bump the `Beam` version on L34 to 2.6.0. And likely the bigqueryVersion which will be a bit more annoying.

PR #134, opened 2018-10-24
Adds support for fixed type.

mfranberg:
* Adds a unit test for all currently supported avro types.
* Makes bytes and fixed consistent with hex strings.
* Fixes incorrect casting for byte and enum types.

### DIFF
--- a/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSamplerAvro.scala
+++ b/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSamplerAvro.scala
@@ -18,6 +18,7 @@
 package com.spotify.ratatool.samplers
 
 import java.io.IOException
+import java.nio.ByteBuffer
 import java.util.{List => JList}
 
 import com.google.common.hash.Hasher
@@ -27,9 +28,9 @@ import com.spotify.scio.ScioContext
 import com.spotify.scio.io.{Tap, Taps}
 import org.apache.avro.Schema
 import org.apache.avro.Schema.Type
-import org.apache.avro.generic.GenericRecord
+import org.apache.avro.generic.{GenericData, GenericFixed, GenericRecord}
+import org.apache.commons.codec.binary.Hex
 import org.slf4j.LoggerFactory
-
 
 import scala.collection.JavaConverters._
 import scala.annotation.tailrec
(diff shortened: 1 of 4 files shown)


ratatool-common/src/test/resources/big_sampler_avro_test_schema.avsc
```diff
@@ -0,0 +1,20 @@
+{
+    "type": "record",
+    "name": "BigSamplerAvroTestRecord",
```
idreeskhan: Can just rename to `FlatAvroSchema` or something that better describes the purpose. Any reason to not just use the `RequiredNestedRecord` in the other one?

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSamplerAvro.scala
```diff
@@ -60,9 +62,10 @@ private[samplers] object BigSamplerAvro {
             v.asInstanceOf[GenericRecord],
             subfields.tail.mkString(BigSampler.fieldSep.toString),
             hasher)
-        case Type.ENUM => hasher.putString(v.asInstanceOf[Enum[_]].name, BigSampler.utf8Charset)
+        case Type.ENUM =>
+          hasher.putString(v.asInstanceOf[EnumSymbol].toString, BigSampler.utf8Charset)
```
idreeskhan: For my own learning, what is the difference here? Avoids a casting error?

ratatool-sampling/src/test/scala/com/spotify/ratatool/samplers/BigSamplerAvroTest.scala
```diff
@@ -0,0 +1,94 @@
...
+import java.nio.ByteBuffer
+import java.util.Arrays
+
+import com.google.common.hash.Hashing
+import com.spotify.ratatool.Schemas
+import org.apache.avro.generic.GenericData.EnumSymbol
+import org.apache.avro.generic.{GenericData, GenericRecordBuilder}
+import org.scalatest.{FlatSpec, Matchers}
+
+class BigSamplerAvroTest extends FlatSpec with Matchers {
```
idreeskhan: We have some property based tests for hasher already defined in here https://github.com/spotify/ratatool/blob/master/ratatool-sampling/src/test/scala/com/spotify/ratatool/samplers/BigSamplerTest.scala#L36

I think it makes sense to consolidate those into the same file, and keep the JobTests in a separate file.

idreeskhan: Also another note outside scope of this PR, would be nice to rethink some of these as property based tests as well 😉 But not a blocker, up to you

ratatool-common/src/test/resources/big_sampler_avro_test_schema.avsc
mfranberg: Good idea I'll use the required ones!

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSamplerAvro.scala
mfranberg: The problem was that you got an EnumSymbol not an Enum and the program failed with an exception. Same happened for Bytes, where I got a ByteBuffer instead of Array[Byte]. Did this work before? :-)

ratatool-sampling/src/test/scala/com/spotify/ratatool/samplers/BigSamplerAvroTest.scala
mfranberg: Fixed!

I agree it would be nice to rethink some of these tests! :-)

mfranberg:
@idreeskhan Moved the tests into the property based testing file. Added tests for generic records to make sure we get the casting right. As a bonus I refactored the BigSamplerAvro by extracting hashing of the primitive types into its own function.

idreeskhan:
This is great, LGTM, I can merge when you are ready @mfranberg

mfranberg:
Squashed history, feel free to merge at your convenience!

PR #135, opened 2018-11-02
Fixes sample fraction logic

mfranberg:
The current implementation of sample fraction takes too many samples when the fraction <= 0.01.


idreeskhan:
Wow fun stuff. LGTM 🎉

Issue #137, opened 2018-11-06
Support multiple keys in BigDiffy


martinbomio:
Right now BigDiffy support a single key. Could be nice to be able to provide multiple keys as a comma serparated list like:
`ratatool bigDiffy --mode avro --key=key_1, key_2`

idreeskhan:
Merged to master, waiting on a release. #141 related to benchmarking

PR #138, opened 2018-11-06
Support multiple keys in bigdiffy

martinbomio:
Addresses #137.
There were two options that I thought about:
1) Join all keys and to get a single String in `*KeyFn` functions.
2) Use the list of string representation of the keys as key to the join.

Decided to go with (1) since we will need to join the keys anyways to write the results.

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -419,7 +419,9 @@ object BigDiffy extends Command {
       } else {
         get(xs, i + 1, r.get(xs(i)).asInstanceOf[GenericRecord])
       }
-    (r: GenericRecord) =>  keys.map { k => get(k.split('.'), 0, r) }.mkString("_")
+
+    val xs = keys.map(_.split('.'))
+    (r: GenericRecord) =>  xs.map { x => get(x, 0, r) }.mkString("_")
   }
 
   private def tableRowKeyFn(keys: Seq[String]): TableRow => String = {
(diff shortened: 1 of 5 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -414,28 +414,27 @@ object BigDiffy extends Command {
...
     @tailrec
     def get(xs: Array[String], i: Int, r: GenericRecord): String =
       if (i == xs.length - 1) {
         r.get(xs(i)).toString
       } else {
         get(xs, i + 1, r.get(xs(i)).asInstanceOf[GenericRecord])
       }
-    val xs = key.split('.')
-    (r: GenericRecord) => get(xs, 0, r)
+    (r: GenericRecord) =>  keys.map { k => get(k.split('.'), 0, r) }.mkString("_")
```
idreeskhan: Hmm, what if we change this to a `Set[String]` instead? My concern is this can produce some weird edge cases. They will probably be rare but it's not really transparent to end user. 

e.g.
`key1_something, key2` vs `key1, something_key2` produces the same aggregate key.

idreeskhan: Only suggesting `Set` since we use that in sampler but open to other ideas

martinbomio: wouldn't it be more confusing is the results don't respect the user ordering of the keys?

idreeskhan:
Can you also add a test that runs with multiple keys?
Also would be good to update the docs with the new format.

idreeskhan:
Sorry, was travelling for a bit. Can you resolve conflicts? and then we can try to get this merged

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
idreeskhan: Maybe some dataset that preserves order instead. Up to you. Doesn't block this PR imo but would be nice to have fixed :)

martinbomio:
@idreeskhan PTAL, rebased master and added multikey case class

idreeskhan:
LGTM, if you have a chance could you do some quick and dirty benchmarking to see the performance impact of boxing? If not we can probably just create an open issue for it and revisit if it crops up.

martinbomio:
@idreeskhan I created https://github.com/spotify/ratatool/issues/141

idreeskhan:
👍

PR #145, opened 2019-02-19
Move parquet to a separate project

jackdingilian:
Move ParquetSampler to a separate `ratatool-extras` project so that we no longer need to pull in the Hadoop dependency when it's not being used.  Remove from the CLI so it doesn't get packaged with the ratatool-cli jar.


anne-decusatis:
I took a look at this. I don't understand how to access the ParquetSampler code in ratatool-extras - an example of how to use this in its new location would be super helpful! (And if there's not a way to use it, we should probably just remove it.)

ratatool-cli/src/main/scala/com/spotify/ratatool/tool/Ratatool.scala
```diff
@@ -63,9 +62,6 @@ object Ratatool {
                 val table = BigQueryIO.parseTableSpec(o.tableOut)
                 BigQueryIO.writeToTable(data, sampler.schema, table)
               }
-            case "parquet" =>
-              val data = new ParquetSampler(new Path(o.in)).sample(o.n, o.head)
```
idreeskhan: We should log something out to the user to raise awareness if people are affected by this move

build.sbt
```diff
@@ -118,11 +118,9 @@ lazy val ratatoolSampling = project
       "org.apache.beam" % "beam-runners-google-cloud-dataflow-java" % beamVersion,
       "com.twitter" %% "algebird-core" % algebirdVersion,
       "joda-time" % "joda-time" % jodaTimeVersion,
-      "org.apache.parquet" % "parquet-avro" % parquetVersion,
       "org.scalacheck" %% "scalacheck" % scalaCheckVersion,
       "org.scalatest" %% "scalatest" % scalaTestVersion % "test",
-      "com.google.cloud.bigdataoss" % "gcs-connector" % gcsVersion,
-      "org.apache.hadoop" % "hadoop-client" % hadoopVersion exclude ("org.slf4j", "slf4j-log4j12")
+      "com.google.cloud.bigdataoss" % "gcs-connector" % gcsVersion
```
idreeskhan: Is this still used?

build.sbt
```diff
@@ -180,6 +178,25 @@ lazy val ratatoolShapeless = project
     ratatoolSampling
   )
 
+lazy val ratatoolExtras = project
```
idreeskhan: Do we need to override some setting here so it does not get `pack`ed?

build.sbt
jackdingilian: It's still used but I should have moved it to `-extras`

jackdingilian:
@anne-decusatis good point I will a readme for `-extras`

build.sbt
jackdingilian: (we discussed offline and this is fine as is)

ratatool-extras/README.md
```diff
@@ -0,0 +1,13 @@
+Parquet Sampler
+===============
+Contains record sampling for Parquet.  Only operates in `head` mode where it samples record from the
+start. We are pulling `ParquetSampler` into it's own project in order to isolate the
```
anne-decusatis: "it's own project"  should be -> "its own project"

ratatool-extras/src/main/scala/com/spotify/ratatool/io/ParquetIO.scala
```diff
@@ -40,7 +22,7 @@ object ParquetIO {
     val reader = AvroParquetReader.builder[T](path)
       .withConf(conf)
       .build()
-      .asInstanceOf[ParquetReader[T]]
+
```
anne-decusatis: was this change intentional?

ratatool-extras/src/main/scala/com/spotify/ratatool/samplers/ParquetSampler.scala
```diff
@@ -30,7 +29,7 @@ import scala.collection.mutable.ListBuffer
  * Records are represented as Avro [[GenericRecord]].
  * Only head mode is supported.
  */
-class ParquetSampler(path: Path, protected val seed: Option[Long] = None)
+class ParquetSampler(path: String, protected val seed: Option[Long] = None)
```
anne-decusatis: this might need to be changed back, per readme example. or the readme example needs to be changed, either one

ratatool-extras/src/main/scala/com/spotify/ratatool/io/ParquetIO.scala
jackdingilian: yeah the `asInstanceOf` wasn't necessary.  Doesn't make a big difference either way though

ratatool-extras/src/main/scala/com/spotify/ratatool/samplers/ParquetSampler.scala
jackdingilian: oops this is correct I will fix the readme

ratatool-sampling/README.md
```diff
@@ -1,7 +1,7 @@
 Sampler
 =======
 
-Diffy contains record sampling classes for Avro, BigQuery, Parquet. Supported filesystems include Local, GCS (`gs://`), and HDFS (`hdfs://`).
```
idreeskhan: Add a note that parquet is supported in extras

PR #148, opened 2019-02-26
Upgrade scio 0.7.1

catherinejelder:
most changes are from the scio 0.7 migration guide: https://spotify.github.io/scio/migrations/v0.7.0-Migration-Guide


project/plugins.sbt
```diff
@@ -4,6 +4,7 @@ addSbtPlugin("com.github.gseitz" % "sbt-protobuf" % "0.6.3")
 addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.7")
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "1.1.0")
 addSbtPlugin("org.scalastyle" % "scalastyle-sbt-plugin" % "1.0.0")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.0")
```
idreeskhan: If you have some time would be nice to bump some of the other plugin versions as well but not necessary. I think scoverage and scalastye have new versions

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -128,7 +131,9 @@ case class FieldStats(field: String,
 
 /** Big diff between two data sets given a primary key. */
 class BigDiffy[T](lhs: SCollection[T], rhs: SCollection[T],
-                  diffy: Diffy[T], keyFn: T => MultiKey) {
+                  diffy: Diffy[T], keyFn: T => MultiKey)
+                 (implicit coder0: Coder[(String, T)],
+                  coder1: Coder[(MultiKey, (String, T))]) {
```
idreeskhan: It should be fine as long as we have a coder for `T` and `MultiKey` right? I think `MultiKey` it should also automatically derive since it's a case class so I would try with just a `T: Coder` context bound

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -175,7 +180,9 @@ object BigDiffy extends Command {
   type DeltaSCollection = SCollection[(MultiKey, (Seq[Delta], DiffType.Value))]
 
   private def computeDeltas[T](lhs: SCollection[T], rhs: SCollection[T],
-                               d: Diffy[T], keyFn: T => MultiKey): DeltaSCollection = {
+                               d: Diffy[T], keyFn: T => MultiKey)
+                              (implicit coder0: Coder[(String, T)],
```
idreeskhan: Same comment here

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -261,15 +268,19 @@ object BigDiffy extends Command {
...
     new BigDiffy[T](lhs, rhs, d, keyFn)
 
   /** Diff two Avro data sets. */
   def diffAvro[T <: GenericRecord : ClassTag](sc: ScioContext,
                                               lhs: String, rhs: String,
                                               keyFn: T => MultiKey,
                                               diffy: AvroDiffy[T],
-                                              schema: Schema = null): BigDiffy[T] =
+                                              schema: Schema = null)
+                                             (implicit coder: Coder[T]): BigDiffy[T] =
```
idreeskhan: This can be `T <: GenericRecord : ClassTag : Coder` instead

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -285,6 +286,9 @@ object BigSampler extends Command {
                                         keyFn: T => U,
                                         maxKeySize: Int = 1e6.toInt,
                                         byteEncoding: ByteEncoding = RawEncoding)
+                                      (implicit coder0: Coder[T],
```
idreeskhan: `U: Coder` and `T: Coder` bounds instead. Potentially can remove the ClassTag bounds now but not certain

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/package.scala
```diff
@@ -55,6 +56,7 @@ package object samplers {
                                                  precision: Precision = Approximate,
                                                  maxKeySize: Int = 1e6.toInt,
                                                  byteEncoding: ByteEncoding = RawEncoding)
+                                               (implicit coder: Coder[T])
```
idreeskhan: Same thing here

project/plugins.sbt
ClaireMcGinty: hey, you should be able to remove this plugin after applying the rules! We should add that to the Scio migration guide...

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/util/SamplerSCollectionFunctions.scala
```diff
@@ -112,6 +120,9 @@ object SamplerSCollectionFunctions {
   private[samplers] def uniformParams[T: ClassTag, U: ClassTag](s: SCollection[T],
                                                                 keyFn: T => U,
                                                                 prob: Double)
+                                                               (implicit coder0: Coder[T],
```
ClaireMcGinty: I think these implicits could be simplified similar to how Scio handles keyed SCollections :)  https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/values/SCollection.scala#L461

catherinejelder:
@idreeskhan there are 2 places still using `implicit coder` param because I couldn't figure out how to convert them to context bounds. suggestions welcome if you have any

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/util/SamplerSCollectionFunctions.scala
catherinejelder: thanks you @ClaireMcGinty! would you mind reviewing the last commit in this PR, "use context bounds to simplify coders"?

catherinejelder: I'm about to merge this but I can incorporate any suggestions later :)

Issue #152, opened 2019-03-26
BigDiffy throws a NPE when outputting to BQ


idreeskhan:
User reported NPE when using ratatool from CLI

```Caused by: java.lang.NullPointerException at com.spotify.ratatool.diffy.BigDiffy$$anonfun$7$$anonfun$apply$4.apply(BigDiffy.scala:337) at com.spotify.ratatool.diffy.BigDiffy$$anonfun$7$$anonfun$apply$4.apply(BigDiffy.scala:330) at scala.Option.map(Option.scala:146) at com.spotify.ratatool.diffy.BigDiffy$$anonfun$7.apply(BigDiffy.scala:330) at com.spotify.ratatool.diffy.BigDiffy$$anonfun$7.apply(BigDiffy.scala:329) at com.spotify.scio.util.Functions$$anon$7.processElement(Functions.scala:145)```

PR #154, opened 2019-04-08
BigDiffy: Delta is serialized, shouldn't contain nulls

anne-decusatis:
Change Delta case class to store LHS and RHS as Options, so we're not shuffling around and eventually saving a null value when one half of the diff is null

This solves an NPE which occurs in Dataflow when diffing two fields, one of which is null and one of which isn't. The NPE occurs at the end of the pipeline, when trying to write diff results to BigQuery. The exception's trace contains `Caused by: org.apache.beam.sdk.coders.CoderException: cannot encode a null value`. 

Fixes https://github.com/spotify/ratatool/issues/152

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -178,7 +178,7 @@ object BigDiffy extends Command {
   type DeltaSCollection = SCollection[(MultiKey, (Seq[Delta], DiffType.Value))]
 
   private def computeDeltas[T : Coder](lhs: SCollection[T], rhs: SCollection[T],
-                               d: Diffy[T], keyFn: T => MultiKey): DeltaSCollection = {
+                                       diffy: Diffy[T], keyFn: T => MultiKey): DeltaSCollection = {
     // extract keys and prefix records with L/R sub-key
     val lKeyed = lhs.map(t => (keyFn(t), ("l", t)))
     val rKeyed = rhs.map(t => (keyFn(t), ("r", t)))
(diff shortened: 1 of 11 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -345,7 +345,10 @@ object BigDiffy extends Command {
               case _ =>
               DeltaValueBigQuery("UNKNOWN", None)
             }
-            DeltaBigQuery(d.field, d.left.toString, d.right.toString, dv)})
```
idreeskhan: I think this is probably fine, but it seems like a lot of code for such a small fix here. Technically our handling of Options is "more" correct, but we are also now modifying the underlying value and taking on responsibility to handle it correctly.

https://github.com/spotify/ratatool/pull/154/files#diff-2a46b923154e83ede309229ad1d38357R326 still passes right?

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -191,15 +191,15 @@ object BigDiffy extends Command {
 
     (lKeyed ++ rKeyed)
       .groupByKey
-      .map { case (k, vs) =>
-        val m = vs.toMap // L/R -> record
-        if (m.size == 2) {
-          val ds = d(m("l"), m("r"))
-          val dt = if (ds.isEmpty) DiffType.SAME else DiffType.DIFFERENT
-          (k, (ds, dt))
+      .map { case (key, values) => // values is a list of tuples: "l" -> record or "r" -> record
```
anne-decusatis: these changes are just making the variable names more readable

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
anne-decusatis: Almost all of the code change is wrapping expected values in tests, if that makes you feel better! And, in my opinion, it was always our responsibility to handle the underlying values correctly. If it wasn't, I don't think this NPE would be a bug to fix :)

https://github.com/anne-decusatis/ratatool/blob/fix_bigdiffy_npe/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/Diffy.scala#L68 via https://github.com/anne-decusatis/ratatool/blob/fix_bigdiffy_npe/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala#L76 should be where that line above is used. so now the values will be wrapped in `Some(...)` but it should still work, I think. I can unwrap them and replace with `""` like the others do if it's a problem, but personally I think it's clearer to not have empty strings passed around. what do you think?

idreeskhan: Should keep the output behaviour consistent if possible. Otherwise, people consuming output for analysis also need to be aware of this change.

anne-decusatis: great, I will update to unwrap in the toString

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -345,7 +345,10 @@ object BigDiffy extends Command {
               case _ =>
               DeltaValueBigQuery("UNKNOWN", None)
             }
-            DeltaBigQuery(d.field, d.left.toString, d.right.toString, dv)})
+            DeltaBigQuery(d.field,
+              d.left.map(_.toString).getOrElse("null"),
```
anne-decusatis: the `getOrElse("null")` is consistent with how nulls were handled before:
```
scala> val s: Any = null                                                                 
val s: Any = null                                                                        
s: Any = null

scala> s"$s"
s"$s"
res0: String = null

scala> res0.length     
res0.length
res1: Int = 4
```

anne-decusatis: (i left this comment in the wrong place but it applies here too)

idreeskhan:
👍 LGTM thanks for updating

Issue #146, opened 2019-02-22
Error when schemas don't match


comeonism:
Hi, 

I've made a change to my avro schema and added a field. Now I'd like to compare the output before the schema change to the output after, to make sure that the rest of the data is still the same. So I do something like this

```
ratatool bigDiffy --region='europe-west1' --numWorkers=60 --maxNumWorkers=100 --experiments=shuffle_mode=service --autoscalingAlgorithm=THROUGHPUT_BASED --workerMachineType=n1-highmem-16 --workerDiskType=compute.googleapis.com/projects//zones//diskTypes/pd-ssd  --runner=DataflowRunner --tempLocation="gs://something/project.fact.gcs.hourly.alex/testing" --input-mode=avro --output-mode=bigquery --key=log_record_id --lhs="gs://jobs/project.fact.gcs.hourly/2019-02-21T13/random-id/*.avro" --rhs="gs://jobs/project.fact.gcs.hourly.new/2019-02-21T13/random-id/*.avro" --output="project:experimental.appidtest" --ignore="app_id"
```

where `rhs` has the new schema. `app_id` is the new field that I've added, that's why I'm putting it into ignore. 

And then I'm getting this exception: 

```
[ForkJoinPool-1-worker-13] ERROR org.apache.beam.runners.dataflow.util.MonitoringUtil$LoggingHandler - 2019-02-22T17:59:28.410Z: java.io.IOException: Failed to start reading from source: gs://jobs/project.fact.gcs.hourly/2019-02-21T13/random-id/part-01749-of-02897.avro range [0, 1234041)
    at com.google.cloud.dataflow.worker.WorkerCustomSources$BoundedReaderIterator.start(WorkerCustomSources.java:595)
    at com.google.cloud.dataflow.worker.util.common.worker.ReadOperation$SynchronizedReaderIterator.start(ReadOperation.java:360)
    at com.google.cloud.dataflow.worker.util.common.worker.ReadOperation.runReadLoop(ReadOperation.java:193)
    at com.google.cloud.dataflow.worker.util.common.worker.ReadOperation.start(ReadOperation.java:158)
    at com.google.cloud.dataflow.worker.util.common.worker.MapTaskExecutor.execute(MapTaskExecutor.java:75)
    at com.google.cloud.dataflow.worker.BatchDataflowWorker.executeWork(BatchDataflowWorker.java:393)
[...]

PR #155, opened 2019-04-09
Assert that schema is backwards compatible for Avro

idreeskhan:
Related to #146 

This assumes that the RHS schema is the "new" schema (backwards compatible to LHS). I also moved this outside of AvroDiffy.

Need to add some tests and docs but curious to hear thoughts on this

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -28,6 +28,7 @@ import com.spotify.scio.values.SCollection
 import com.spotify.scio.coders.Coder
 import com.twitter.algebird._
 import org.apache.avro.Schema
+import org.apache.avro.SchemaValidatorBuilder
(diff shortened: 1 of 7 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -489,11 +490,14 @@ object BigDiffy extends Command {
...
     val result = inputMode match {
       case "avro" =>
-        // TODO: handle schema
-        val schema = new AvroSampler(rhs, conf = Some(sc.options))
+        val schemaRhs = new AvroSampler(rhs, conf = Some(sc.options))
           .sample(1, head = true).head.getSchema
+        val schemaLhs = new AvroSampler(lhs, conf = Some(sc.options))
+          .sample(1, head = true).head.getSchema
+        new SchemaValidatorBuilder().canReadStrategy.validateLatest()
+          .validate(schemaRhs, Seq(schemaLhs).toList.asJava)
```
anne-decusatis: nitpick: can initialize as a List instead of calling Seq.toList

Issue #157, opened 2019-04-11
distributionFields on null avro field - null exception


artyomboyko:
I've tried to sample data on `distributionFields` which can be null.

`ratatool bigSampler --sample=0.01 --distribution=stratified --distributionFields=someField --input=... --output=... --project=... --region=europe-west1 --tempLocation=... --runner=DataflowRunner --serviceAccount=...`

First I've got null exception here https://github.com/spotify/ratatool/blob/master/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSamplerAvro.scala#L77

After adding try {} catch {} to that code and passing `--nullableCoders=true` to cli, I'm still getting null exception, this time from scio coders.

This might be related: https://github.com/spotify/scio/pull/1820

```
2019-04-10T14:44:30.173Z: java.lang.IllegalArgumentException: Unable [331/1394]
element 'Set(null)' with coder 'SetCoder'.
        at org.apache.beam.sdk.coders.Coder.getEncodedElementByteSize(Coder.java:300)
        at org.apache.beam.sdk.coders.Coder.registerByteSizeObserver(Coder.java:291)
        at com.spotify.scio.coders.instances.BaseSeqLikeCoder.registerByteSizeObserver(ScalaCoders.scala:66)
        at org.apache.beam.sdk.coders.NullableCoder.registerByteSizeObserver(NullableCoder.java:142)
        at com.spotify.scio.coders.WrappedBCoder.registerByteSizeObserver(Coder.scala:217)
        at com.spotify.scio.coders.RecordCoder.registerByteSizeObserver(Coder.scala:333)
        at com.spotify.scio.coders.WrappedBCoder.registerByteSizeObserver(Coder.scala:217)                                                                                                   at org.apache.beam.sdk.util.WindowedValue$FullWindowedValueCoder.registerByteSizeObserver(WindowedValue.java:564)
        at org.apache.beam.sdk.util.WindowedValue$FullWindowedValueCoder.registerByteSizeObserver(WindowedValue.java:480)
        at org.apache.beam.runners.dataflow.worker.IntrinsicMapTaskExecutorFactory$ElementByteSizeObservableCoder.registerByteSizeObserver(IntrinsicMapTaskExecutorFactory.java:400)
        at org.apache.beam.runners.dataflow.worker.util.common.worker.OutputObjectAndByteCounter.update(OutputObjectAndByteCounter.java:125)
[...]

PR #158, opened 2019-04-12
Fix BigSampler NPE

anne-decusatis:
Fixes https://github.com/spotify/ratatool/issues/157

Neville suggested that we avoid trying to serialize `Object`/`Any`/`AnyRef` because we can't serialize those at compile time. Moving the `Set[Any]` key to a `Set[String]` means that we don't even need the nullableCoders fixes in Scio 0.8.0 for it to work :)

### DIFF
--- a/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
+++ b/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
@@ -360,7 +360,6 @@ object BigSampler extends Command {
   def run(argv: Array[String]): Unit = {
     this.singleInput(argv)
   }
-
 }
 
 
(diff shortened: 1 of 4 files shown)


ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -360,6 +360,7 @@ object BigSampler extends Command {
   def run(argv: Array[String]): Unit = {
     this.singleInput(argv)
   }
+
```
idreeskhan: Extreme nitpick, spurious line add

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSamplerBigQuery.scala
```diff
@@ -120,10 +120,11 @@ private[samplers] object BigSamplerBigQuery {
 
   private[samplers] def buildKey(schema: => Seq[TableFieldSchema],
                                  distributionFields: Seq[String])(tr: TableRow)
-  : Set[Any] = {
+  : Set[String] = {
     distributionFields.map{ f =>
       getTableRowField(tr, f, schema)
-    }.toSet
+    }
+      .map(_.toString).toSet
```
idreeskhan: Nitpick, why not bump this up to previous line?

Issue #90, opened 2018-05-10
Generators should work for `arbitrary[T]`


idreeskhan:
Should be fairly simple for the existing Avro, TableRow, Protobuf generators. Either verify if they work already of implement if not.

PR #159, opened 2019-04-30
Arbitrary instances for SpecificRecord and Protobuf

idreeskhan:
#90


ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/package.scala
```diff
@@ -165,4 +168,19 @@ package object scalacheck extends AvroGeneratorOps
...
+  }
+
+  def arbTableRow(schema: TableSchema): Arbitrary[TableRow] = {
+    Arbitrary(tableRowOf(schema))
+  }
+
+  def arbGenericRecord(schema: Schema): Arbitrary[GenericRecord] = {
+    Arbitrary(genericRecordOf(schema))
+  }
+}
```
anne-decusatis: scalastyle is failing on the EOL not being a newline

ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/package.scala
```diff
@@ -165,4 +168,19 @@ package object scalacheck extends AvroGeneratorOps
     }
   }
 
-}
+  implicit def arbSpecificRec[T <: SpecificRecord : ClassTag]: Arbitrary[T] = {
```
anne-decusatis: do we want to document these in a README?

ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/package.scala
```diff
@@ -165,4 +168,19 @@ package object scalacheck extends AvroGeneratorOps
...
+    Arbitrary(specificRecordOf[T])
+  }
+
+  implicit def arbProtobuf[T <: AbstractMessage : ClassTag]: Arbitrary[T] = {
```
anne-decusatis: I don't have a clear sense of the issue and there's not much description in https://github.com/spotify/ratatool/issues/90 so I'm not sure why these are implicit. Do they need to be?

idreeskhan: Yes since Arbitrary is a type class for implicitly finding generators. I only added the other explicit ones for completeness sake but they could be excluded

ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/package.scala
idreeskhan: 👍 Added a note

PR #160, opened 2019-05-20
Allow bigDiffy to work with nullable keys

andrusha:
Comparing datasets which were grouped by compound key might result in one of the keys having `null` value.

Eg: we have a table grouped by `(company, country, product)` for total number of sales. It's possible that one of the products is `null` as there is no single unique key in this table in order to compare two versions I have to create compound key with `--key=company --key=country --key=product` as `product` might be `null` the unpatched version would throw `NullPointerException`:

```
[ForkJoinPool-1-worker-13] ERROR org.apache.beam.runners.dataflow.util.MonitoringUtil$LoggingHandler - 2019-05-17T14:33:47.344Z: java.lang.NullPointerException
    at com.spotify.ratatool.diffy.BigDiffy$.com$spotify$ratatool$diffy$BigDiffy$$get$1(BigDiffy.scala:428)
    at com.spotify.ratatool.diffy.BigDiffy$$anonfun$avroKeyFn$1$$anonfun$apply$111.apply(BigDiffy.scala:434)
    at com.spotify.ratatool.diffy.BigDiffy$$anonfun$avroKeyFn$1$$anonfun$apply$111.apply(BigDiffy.scala:434)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
    at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:234)
    at scala.collection.immutable.List.foreach(List.scala:392)
    at scala.collection.TraversableLike$class.map(TraversableLike.scala:234)
    at scala.collection.immutable.List.map(List.scala:296)
    at com.spotify.ratatool.diffy.BigDiffy$$anonfun$avroKeyFn$1.apply(BigDiffy.scala:434)
    at com.spotify.ratatool.diffy.BigDiffy$$anonfun$avroKeyFn$1.apply(BigDiffy.scala:434)
    at com.spotify.ratatool.diffy.BigDiffy$$anonfun$63.apply(BigDiffy.scala:183)
    at com.spotify.ratatool.diffy.BigDiffy$$anonfun$63.apply(BigDiffy.scala:183)
    at com.spotify.scio.util.Functions$$anon$7.processElement(Functions.scala:172)
```

PR #161, opened 2019-05-28
Warn on null keys

jackdingilian:
https://github.com/spotify/ratatool/pull/160 allowed for null values in keys.  This adds a warning whenever we encounter a null key.  Also throw an error whenever keys are non-unique as that might be somewhat easier to do by accident for nullable fields now

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -194,17 +194,18 @@ object BigDiffy extends Command {
     (lKeyed ++ rKeyed)
       .groupByKey
       .map { case (key, values) => // values is a list of tuples: "l" -> record or "r" -> record
+        if (values.size > 2) {
+          throw new RuntimeException(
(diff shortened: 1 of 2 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -429,7 +436,13 @@ object BigDiffy extends Command {
     @tailrec
     def get(xs: Array[String], i: Int, r: GenericRecord): String =
       if (i == xs.length - 1) {
-        String.valueOf(r.get(xs(i)))
+        val valueOfKey = r.get((xs(i)))
+        if (valueOfKey == null) {
+          logger.warn(
+            s"""Null value found for key: ${xs.mkString(".")}.
+               | If this is not expected check your or use a different key.""".stripMargin)
```
catherinejelder: "check your key" instead of "check your"?

jackdingilian: oops thanks should be "check your data"

PR #162, opened 2019-08-16
add check that GCS output begins with gs://

anne-decusatis:
We received a user report that Ratatool succeeded in running a diff but they didn't know where the result was output to. We should fail fast if the output location is invalid.


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -504,6 +505,13 @@ object BigDiffy extends Command {
       case m => throw new IllegalArgumentException(s"output mode $m not supported")
     }
 
+    if (om == GCS && !output.startsWith("gs://")) {
```
anne-decusatis: is it in scope to fail fast similarly for BQ?

PR #165, opened 2019-08-20
Move from Travis CI to Circle CI

anne-decusatis:
based on bits of Scio's config.yml

PR to test review builds :)


anne-decusatis:
oh, also note that I ignored/didn't even try to copy the deploy section of .travis.yml because we don't use it https://github.com/spotify/ratatool/blob/master/.travis.yml#L32

PR #166, opened 2019-08-21
Add namespace to TestRecord enum & fixed (fix compile in circleci?)

anne-decusatis:
Sometimes these are generated without a namespace
Then, we fail to compile because of this Scala bug: https://github.com/scala/bug/issues/4071?orig=1
We can make sure it gets generated with a namespace by specifying namespace
this is valid avro per https://avro.apache.org/docs/current/spec.html#Enums


anne-decusatis:
merging this myself because I meant to put these on the same branch & couldn't figure out how to get git working appropriately without this PR :(

PR #169, opened 2019-09-19
Add ability to derive Gen for case classes

anne-decusatis:
Given a case class containing fields, and (potentially non-arbitrary) generators for individual 
fields in the implicit scope, generates a Gen for the case class.


build.sbt
```diff
@@ -34,6 +34,7 @@ val bigqueryVersion = "v2-rev374-1.23.0"
 val beamVersion = "2.10.0"
 val guavaVersion = "25.1-jre"
 val shapelessVersion = "2.3.3"
+val magnoliaVersion = "0.10.1-jto" // needs to stay in sync with scio version
```
jackdingilian: scio is using `0.11.0` now https://github.com/spotify/scio/blob/master/build.sbt#L72

anne-decusatis: @jackdingilian that change was made last week, I'm not sure it's released yet (https://github.com/spotify/scio/pull/2241/files)

There's a change between 0.11 and 0.10 in Magnolia that causes a pipeline using both scio and ratatool-scalacheck to fail compilation with `java.lang.NoSuchMethodError: magnolia.TypeName.<init>` if scio and ratatool aren't on the same Magnolia version; see also the breaking change PR in question: https://github.com/propensive/magnolia/pull/152

Maybe I should call this out in the README better?

jackdingilian: oh cool nvm then

anne-decusatis: updated README, thanks

PR #202, opened 2019-12-03
Adding flag to ignore NaN values

danielblazevski:


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -483,14 +491,16 @@ object BigDiffy extends Command {
 
   /** Scio pipeline for BigDiffy. */
   //scalastyle:off cyclomatic.complexity
+  //scalastyle:off method.length
```
catherinejelder: nit: it's common to put all scalastyle configs on one line like //scalastyle:off cyclomatic.complexity method.length

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -483,14 +491,16 @@ object BigDiffy extends Command {
...
     val (sc, args) = ContextAndArgs(cmdlineArgs)
 
-    val (inputMode, keys, lhs, rhs, output, header, ignore, unordered, outputMode) = {
+    val (inputMode, keys, lhs, rhs, output, header, ignore, unordered, outputMode, ignoreNan) = {
       try {
         (args("input-mode"), args.list("key"), args("lhs"), args("rhs"), args("output"),
           args.boolean("with-header", false), args.list("ignore").toSet,
-          args.list("unordered").toSet, args.optional("output-mode"))
+          args.list("unordered").toSet, args.optional("output-mode"),
+        args.boolean("ignore-nan", false))
```
catherinejelder: nit: missing an indent?

ratatool-diffy/src/test/scala/com/spotify/ratatool/diffy/BigDiffyTest.scala
```diff
@@ -120,6 +120,32 @@ class BigDiffyTest extends PipelineSpec {
     }
   }
 
+  it should "work with NaNs" in {
```
catherinejelder: can we also test for when `ignoreNan` is false (the default)?

catherinejelder: nit: more descriptive name(s) for the tests would also be good

danielblazevski: Cool, good call.  Added another test, but couldn't verify min/max since how Algebird computes min/maxes when NaN values are present will result in flaky tests - added a comment about this in the new test in 3c81e32

PR #264, opened 2020-04-20
Bump Scio to 0.9.0

nevillelyh:


nevillelyh:
Build is breaking since we dropped scala 2.11 & added 2.13. You need to update `crossScalaVersions` accordingly.

PR #267, opened 2020-04-21
Stringify byte buffers in diffy avro keys

anne-decusatis:
The default toString for a byte buffer doesn't contain its content,
 just metadata that's the same for every byte buffer of the same size.
If you used a bytes field in Avro as your key,
this would cause a hot key of that default string, rather than grouping by key values.

This fix uses the hex version of what's actually inside the buffer.

confirmed with user that diff job succeeds with this change in place

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -21,6 +21,7 @@ import java.nio.ByteBuffer
 import java.nio.charset.StandardCharsets
 
 import com.google.api.services.bigquery.model.{TableFieldSchema, TableRow, TableSchema}
+import com.google.common.io.BaseEncoding
 import com.google.protobuf.AbstractMessage
 import com.spotify.ratatool.Command
 import com.spotify.ratatool.samplers.AvroSampler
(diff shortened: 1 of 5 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -17,7 +17,11 @@
 
 package com.spotify.ratatool.diffy
 
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
```
anne-decusatis, line 21: unused import

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -451,7 +454,13 @@ object BigDiffy extends Command with Serializable {
             s"""Null value found for key: ${xs.mkString(".")}.
                | If this is not expected check your data or use a different key.""".stripMargin)
         }
-        String.valueOf(valueOfKey)
+        // handle bytes keys custom, so we get bytebuffer actual content and not toString metadata
+        if (valueOfKey.isInstanceOf[ByteBuffer]) {
```
jackdingilian, line 458: nit: make this an else if just for readability?

anne-decusatis: ir's not actually an else-if here, the if above just logs a warning and continues to return what's beneath it

PR #269, opened 2020-04-27
bump to scio 0.9

danielblazevski:
Main changes are dropping scala 2.11 and sc.avroFile now has SpecificRecord as a type bound
https://github.com/spotify/scio/blob/4609bd7a2fe0359ea45e60caa29e909073b9add7/scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala#L82

Unless the type param is unspecified, as here
https://github.com/spotify/scio/blob/4609bd7a2fe0359ea45e60caa29e909073b9add7/scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala#L46


build.sbt
```diff
@@ -39,8 +39,8 @@ val commonSettings = Sonatype.sonatypeSettings ++ releaseSettings ++ Seq(
   organization := "com.spotify",
   name := "ratatool",
   description := "A tool for random data sampling and generation",
-  scalaVersion := "2.11.12",
-  crossScalaVersions := Seq("2.11.12", "2.12.10"),
+  scalaVersion := "2.12.10",
+  crossScalaVersions := Seq("2.12.10"),
```
idreeskhan, line 43: We should start compiling for 2.13 as well

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -283,13 +283,13 @@ object BigDiffy extends Command with Serializable {
     new BigDiffy[T](lhs, rhs, d, keyFn, ignoreNan)
 
   /** Diff two Avro data sets. */
-  def diffAvro[T <: GenericRecord : ClassTag : Coder](sc: ScioContext,
+  def diffAvro(sc: ScioContext,
                                               lhs: String, rhs: String,
-                                              keyFn: T => MultiKey,
-                                              diffy: AvroDiffy[T],
+                                              keyFn: GenericRecord => MultiKey,
```
idreeskhan, line 288: What's up with this change?

build.sbt
idreeskhan: Hmm nvm it looks like scio is only compiling for 2.12 rn

idreeskhan: Actually they are, its just specified separately, likely differs by module. https://github.com/spotify/scio/blob/master/build.sbt#L441

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
danielblazevski: type parameters for sc.avroFile now have SpecificRecord as a type bound

https://github.com/spotify/scio/blob/4609bd7a2fe0359ea45e60caa29e909073b9add7/scio-avro/src/main/scala/com/spotify/scio/avro/syntax/ScioContextSyntax.scala#L82

I had no choice but to remove the type parameters for sc.avroFile[GenericRecord], and was 80% sure this would be OK to have keyFn and and diffy not be generic types.

build.sbt
danielblazevski: 👍 will add

danielblazevski: lol yeah. I first thought it was only 2.12 as well.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
anne-decusatis: is this ok if it's being used directly in a pipeline?

idreeskhan: We can retain the type safety here for compatibility but change how we call it in scio. GenericRecord keyFn is also likely much weirder to deal with

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -537,7 +538,10 @@ object BigDiffy extends Command with Serializable {
           .sample(1, head = true).head.getSchema
         implicit val grCoder: Coder[GenericRecord] = Coder.avroGenericRecordCoder(schema)
         val diffy = new AvroDiffy[GenericRecord](ignore, unordered)
-        BigDiffy.diffAvro[GenericRecord](sc, lhs, rhs, avroKeyFn(keys), diffy, schema, ignoreNan)
+        val lhsSCollection = sc.avroFile(lhs, schema)
+        val rhsSCollection = sc.avroFile(rhs, schema)
+        BigDiffy
+          .diff[GenericRecord](lhsSCollection, rhsSCollection, diffy, avroKeyFn(keys), ignoreNan)
```
anne-decusatis, line 544: was it intentional to call .diff and not .diffAvro here? I'm not entirely sure what the impact of this change is on the call stack, can you help walk me through it?

jackdingilian: The scio-avro implementation changed slightly so that `avroFile` either takes a type param that is a sub-type of `SpecificRecord` or no type param and returns `GenericRecord`s.  In order to preserve type safety for users in `diffAvro` we make that operate on SpecificRecords.  So here we need to bypass it and just use diff since we can only operate on `GenericRecord`s from the cli.

PR #272, opened 2020-05-05
scala 2.13 upgrade

jackdingilian:


ratatool-sampling/src/main/scala/com/spotify/ratatool/io/BigQueryIO.scala
```diff
@@ -84,15 +84,15 @@ object BigQueryIO {
 
 }
 
-private class TableRowIterator(private val iterator: PatchedBigQueryTableRowIterator)
+private class TableRowIterator(private val iter: PatchedBigQueryTableRowIterator)
```
catherinejelder, line 87: curious why the rename?

jackdingilian: yeah this is a bit weird.  2.13 collections update added a final field called iterator to Iterators: https://github.com/scala/scala/blob/v2.13.2/src/library/scala/collection/Iterator.scala#L96

So this is now trying to override that and can't.  This maintains compatibility for both 2.12 and 2.13

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -144,6 +143,8 @@ class BigDiffy[T : Coder](@transient val lhs: SCollection[T], @transient val rhs
   private lazy val globalAndFieldStats: SCollection[(GlobalStats, Iterable[FieldStats])] =
     BigDiffy.computeGlobalAndFieldStats(_deltas, ignoreNan)
 
+
+  implicit val deltasCoder: Coder[(MultiKey, String, Any, Any)] = Coder.kryo
```
idreeskhan, line 147: We should add a comment here about using kryo. This might have to change in a future version

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -311,7 +312,7 @@ object BigDiffy extends Command with Serializable {
   /** Merge two BigQuery TableSchemas. */
   def mergeTableSchema(x: TableSchema, y: TableSchema): TableSchema = {
     val r = new TableSchema
-    r.setFields(mergeFields(x.getFields.asScala, y.getFields.asScala).asJava)
+    r.setFields(mergeFields(x.getFields.asScala.toList, y.getFields.asScala.toList).asJava)
```
idreeskhan, line 315: Why this change? Iterator laziness?

ratatool-sampling/src/test/scala/com/spotify/ratatool/io/FixRandomData.scala
```diff
@@ -20,7 +20,7 @@ package com.spotify.ratatool.io
 import com.spotify.ratatool.avro.specific.TestRecord
 
 import java.util.{HashMap => JHashMap, Map => JMap}
-import scala.collection.JavaConversions._
+import scala.jdk.CollectionConverters._
```
idreeskhan, line 23: Should we migrate all instance of `scala.collection.JavaConverters` to this as well? That's the deprecated one now right?

ratatool-shapeless/src/main/scala/com/spotify/ratatool/shapeless/CaseClassDiffy.scala
```diff
@@ -43,7 +43,7 @@ object MapEncoder {
   MapEncoder[FieldType[K, String]] = {
     val name = witness.value.name
     createEncoder { v =>
-      Map(name -> v.self)
+      Map(name -> v.toString)
```
idreeskhan, line 46: What was the issue here?

jackdingilian: In 212 this is a `WrappedString` (https://github.com/scala/scala/blob/v2.12.10/src/library/scala/collection/immutable/WrappedString.scala#L34) 

In 2.13 it is just a `String` so `.self` won't compile.

`toString` in 2.12 returns self anyways so this is equivalent in 2.12 and doesn't break in 2.13.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
jackdingilian: This expects a `Seq` which in 2.13 is an `scala.collection.immutable.Seq`.  `asScala` returns a `mutable.Buffer[TableFieldSchema]` which extends `mutable.Seq`.  It worked in 2.12 because the default Seq was `scala.collection.Seq` which they both inherit from.  Using toList instead of toSeq so it isn't seen as a redundant method call in 2.12

Issue #103, opened 2018-06-11
BigDiffy should support `unorderedKeys` from CLI


idreeskhan:
Bit complicated atm since it expects a map of Key to String. Important thing is having a clean CLI while retaining the current behaviour in BigDiffy.

PR #330, opened 2020-10-06
Fix some warnings and add command line parsing for unordered field keys

idreeskhan:
Don't know if this is a good idea or not but it was annoying to me to have to import ratatool as a dependency just to use this feature (and therefore commit diff jobs into a pipeline repo). Basically allows `->` demarcated field key mappings for nested records that can be ',' delimited to specify multiple field mappings. Let me know what you think

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -431,8 +431,8 @@ object BigDiffy extends Command with Serializable {
         |  --output=<output>                File path prefix for output
         |  --ignore=<keys>                  ',' separated field list to ignore
         |  --unordered=<keys>               ',' separated field list to treat as unordered
-        |  --unorderedFieldKey=<key>        ',' separated list of keys for fields which are unordered nested records. Mappings use '->'
-        |                                   For example --unorderedFieldKey=fieldPath->fieldKey,otherPath->otherKey
+        |  --unorderedFieldKey=<key>        ',' separated list of keys for fields which are unordered nested records. Mappings use ':'
+        |                                   For example --unorderedFieldKey=fieldPath:fieldKey,otherPath:otherKey
         |  [--with-header]                  Output all TSVs with header rows. Defaults to false
         |  [--ignore-nan]                   Ignore NaN values when computing stats for differences
         |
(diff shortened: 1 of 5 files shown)


idreeskhan:
This fixes #103

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -423,12 +425,14 @@ object BigDiffy extends Command with Serializable {
...
         |  --input-mode=(avro|bigquery)     Diff-ing Avro or BQ records
         |  [--output-mode=(gcs|bigquery)]   Saves to a text file in GCS or a BigQuery dataset. Defaults to GCS
-        |  --key=<key>                      '.' separated key field. Specify multiple --key params for multi key usage.
+        |  --key=<key>                      '.' separated key field. Specify multiple --key params or multiple ',' separated key fields for multi key usage.
         |  --lhs=<path>                     LHS File path or BigQuery table
         |  --rhs=<path>                     RHS File path or BigQuery table
         |  --output=<output>                File path prefix for output
         |  --ignore=<keys>                  ',' separated field list to ignore
         |  --unordered=<keys>               ',' separated field list to treat as unordered
+        |  --unorderedFieldKey=<key>        ',' separated list of keys for fields which are unordered nested records. Mappings use '->'
```
idreeskhan, line 434: Thinking of switching this to `:` delimited

anne-decusatis: I like : better than ->

ratatool-diffy/src/test/scala/com/spotify/ratatool/diffy/BigDiffyTest.scala
```diff
@@ -240,6 +242,15 @@ class BigDiffyTest extends PipelineSpec {
     keyValues.toString shouldBe "foo_bar"
   }
 
+  "BigDiffy unorderedKeysMap" should "work with multiple unordered keys" in {
+    val keyMappings = List("record.nested_record->key", "record.other_nested_record->other_key")
+    val unorderdKeys = BigDiffy.unorderedKeysMap(keyMappings)
```
anne-decusatis, line 247: nitpick: typo, unordered

anne-decusatis (approved):
just to confirm, this was already available in scala, we're just adding to CLI? if so, looks good, though I prefer the different delimiter

PR #369, opened 2021-01-20
clarify that bq can be used in cli

catherinejelder:


idreeskhan (approved):
Thanks, you should retrigger the sampling run on circle CI, it is a bit flaky cause of the randomness

PR #410, opened 2021-04-07
[WIP] clarify limits of schema evolution

catherinejelder:
a user ran into an avro error diffing two schemas where the rhs one added a non-nullable field. waiting on her to confirm that a backwards-compatible schema works.
this is based on my understanding of the current behavior after reading [AvroDiffy](https://github.com/spotify/ratatool/blob/master/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala).
not sure yet if we want to change this behavior.

### DIFF
--- a/ratatool-diffy/README.md
+++ b/ratatool-diffy/README.md
@@ -65,8 +65,9 @@ If you need to use `unorderedFieldKeys` to diff nested repeated records, then yo
  as this is currently unsupported from the CLI.
 
 ## Schema Evolution
-If you are trying to diff two schemas that are backwards compatible, you should put the "new" schema
- which is backwards compatible on the RHS. You can also add the new fields to the ignore list to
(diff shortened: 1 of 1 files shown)


idreeskhan (approved):
Don't know if we _can_ change the behaviour. Curious what you had in mind. Thanks for updating!

catherinejelder:
I thought there might be a way to remove the ignored fields from the schema(s) before validating, but I hadn't looked into it and guessed it prob wasn't worth the effort

idreeskhan:
It's not worth the risk of edge cases IMO. We risk having to deal with our own schema validation logic if we do that. The validation is here https://github.com/spotify/ratatool/blob/master/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala#L34

PR #472, opened 2021-08-27
Add option to convert nulls for BQ keys

jamesmcm:
Adds the `convert-nulls` option to convert nulls to "null" for use as keys in BigQuery diffs. This is very useful for comparing output
datasets when `null` is an expected value/category for a string key e.g. in datasets where `null` country codes might be expected for example.

From the code it seems this is already the behaviour in Avro diffs, it is only BQ diffs that would fail previously.

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -484,7 +484,16 @@ object BigDiffy extends Command with Serializable {
     @tailrec
     def get(xs: Array[String], i: Int, r: java.util.Map[String, AnyRef]): String =
       if (i == xs.length - 1) {
-        r.get(xs(i)).toString
+        val valueOfKey = r.get(xs(i))
+        if (valueOfKey == null) {
+          logger.warn(
+            s"""Null value found for key: ${xs.mkString(".")}.
+               | If this is not expected check your data or use a different key.""".stripMargin)
+        }
+
+        // Implicitly converts nulls to "null"
+        // Same as for Avro string case above
+        String.valueOf(valueOfKey)
       } else {
         get(xs, i + 1, r.get(xs(i)).asInstanceOf[java.util.Map[String, AnyRef]])
       }


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -434,6 +434,7 @@ object BigDiffy extends Command with Serializable {
         |  --unorderedFieldKey=<key>        ',' separated list of keys for fields which are unordered nested records. Mappings use ':'
         |                                   For example --unorderedFieldKey=fieldPath:fieldKey,otherPath:otherKey
         |  [--with-header]                  Output all TSVs with header rows. Defaults to false
+        |  [--convert-nulls]                Convert nulls to "null" for use in key fields. Defaults to false
```
idreeskhan, line 437: We should match existing behaviour with Avro and just have this enabled in the lib instead of being a CLI arg. We can update internal docs/SO afterwards to reflect this.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -480,11 +481,27 @@ object BigDiffy extends Command with Serializable {
...
+        // Note this can also be done via Java's String.valueOf(_)
+
+        val keyValue: Option[String] = Option(r.get(xs(i))).map(_.toString)
+
+        (keyValue, convertNulls) match {
+          case (Some(value), _) => value
+          case (None, true) => "null"
+          case (None, false) => {
+          logger.error(
+            s"""Null value found for key: ${xs.mkString(".")}.
```
idreeskhan, line 500: Should match the existing behaviour with Avro. https://github.com/spotify/ratatool/blob/master/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala#L465

idreeskhan: Probably can reuse most of the existing avro code

idreeskhan (commented):
Thanks for the contribution! Would like a few changes. Should also look into why the tests are failing and fix that if possible.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
jamesmcm: Do you also want to warn here? I was just thinking in our case it was like a valid category, so we had hundreds of thousands of rows like that (thus wanting to make it a command line argument before).

jamesmcm:
Btw I had some issues with the formatting (and there are still some small differences vs. before - is there a scalafmt config for the repo?

And do you want me to squash the commits when it is ready to merge?

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
idreeskhan: Yea I think it makes sense to warn here as well, I'd rather err on the side of being too noisy first and course correct later if it's a problem

idreeskhan:
We don't have a scalafmt config right now but you can probably re-use the one from scio in the meantime, we follow the same style guide.

Squashing is preferred

jamesmcm:
Fixed, btw using the scalafmt config from the golden path:

```
version = "2.4.2"
maxColumn = 100

binPack.literalArgumentLists = true

continuationIndent {
  callSite = 2
  defnSite = 2
}

newlines {
  afterImplicitKWInVerticalMultiline = true
  beforeImplicitKWInVerticalMultiline = true
  sometimesBeforeColonInMethodReturnType = true
}

docstrings = JavaDoc

project.git = false

rewrite {
  rules = [PreferCurlyFors, RedundantBraces, RedundantParens, SortImports]
  redundantBraces.maxLines = 1
}
```

Resulted in a lot of changes though so for now I just fixed it manually.

idreeskhan:
Yea that's alright, I can address that separately afterwards

idreeskhan (approved):

PR #473, opened 2021-09-03
Bump Scio to 0.11.0

stormy-ua:


clairemcginty:
hi! I just wanted to leave some context for whoever works on/merges this PR.

I've mentioned the Avro encoding issue from Beam 2.31.0 (`CharSequence` fields are backed by non-serializable `org.apache.avro.util.Utf8`s by default rather than `java.lang.String`s). We weren't able to get a fix for this into the recently released Beam 2.32.0. So, Scio 0.11.0 includes a shaded [AvroCoder](https://github.com/spotify/scio/blob/main/scio-core/src/main/java/org/apache/beam/sdk/coders/shaded/AvroCoder.java#L156) that contains a fix. So, if you rely on a `java.lang.String` representation in your Avro SpecificRecords, you'll have to replace your `org.apache.beam.sdk.coders.AvroCoder.of(classOf[SomeAvroType])` references with `org.apache.beam.sdk.coders.shaded.AvroCoder.of(classOf[SomeAvroType], true)`. The boolean flag tells the coder to use Beam's Reflect encoding/decoding API, which decodes to `java.lang.String`s by default. This class is included in the `scio-core` artifact.

As an alternative, I documented [here](https://spotify.github.io/scio/extras/Sort-Merge-Bucket.html#avro-string-keys) how to modify your Avro schemas to explicitly use a String representation by setting a certain field property. This requires no change to your AvroCoders.

When Beam 2.33.0 is released we can remove the shaded copy in Scio & go back to using the official Beam one.

ratatool-scalacheck/src/main/java/org/apache/beam/sdk/coders/shaded/ScioAvroCoder.java
```diff
@@ -0,0 +1,799 @@
...
+ * the criteria of {@link Coder#verifyDeterministic} by inspecting both the type and the Schema
+ * provided or generated by Avro. Only coders that are deterministic can be used in {@link
+ * org.apache.beam.sdk.transforms.GroupByKey} operations.
+ *
+ * @param <T> the type of elements handled by this coder
+ */
+@SuppressWarnings({
+    "nullness" // TODO(https://issues.apache.org/jira/browse/BEAM-10402)
+})
+public class ScioAvroCoder<T> extends CustomCoder<T> {
```
idreeskhan, line 119: Copying this over from scio so we don't have to add `scio-core` as a dep for generators.

ratatool-scalacheck/src/main/java/org/apache/beam/sdk/coders/shaded/ScioAvroCoder.java
```diff
@@ -0,0 +1,799 @@
...
+import org.apache.beam.sdk.values.TypeDescriptor;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Supplier;
+import org.apache.beam.vendor.guava.v26_0_jre.com.google.common.base.Suppliers;
+import org.checkerframework.checker.nullness.qual.Nullable;
+import org.joda.time.DateTime;
+import org.joda.time.DateTimeZone;
+
+/**
+ *
+ * @todo NOTE: SHADED FROM apache/beam@9375df7. REMOVE WHEN BEAM 2.33.0 IS RELEASED.
```
jackdingilian, line 76: Maybe open an issue to remind us to delete this when scio w 2.33.0 is ready?

jackdingilian (approved):

PR #484, opened 2021-09-30
Scalafmt entire repo

idreeskhan:
* add `.scalafmt.conf`
* run scalafmt on all files
* small refactors to avoid scalastyle error (`BigDiffy.scala`)
* some scalastyle ignores to avoid larger refactors


jackdingilian (approved):
LGTM. I really only reviewed the scalafmt settings and BigDiffy.scala since it sounds like the rest is auto-format but that all looks good

PR #485, opened 2021-10-01
add support for time-partitioned BQ partitions in bigDiffy [WIP]

jrmcglynn:
This PR adds support for comparing natively-partitioned BigQuery tables in `bigDiffy`. Users can add the `$<partition>` decorator to the `lhs` or `rhs` argument. In that case, `bigDiffy` determines the partitioning field and reads the input table from storage with a `rowRestriction` for the partition value.

As is, this only supports time-partitioned fields. Support for range partitioning should be able to be added easily as well.

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -337,23 +337,37 @@ object BigDiffy extends Command with Serializable {
   ): BigDiffy[T] =
     diff(sc.protobufFile(lhs), sc.protobufFile(rhs), diffy, keyFn)
 
+  def isNativelyPartitioned(table: String): Boolean = {
+    // natively partitioned tables should be input as `project:dataset.table$<partition>`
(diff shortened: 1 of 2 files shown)


jrmcglynn:
@catherinejelder @idreeskhan FYI here is a first draft of this.

I got a Dataflow job to start but it did not read any records and failed with `"java.util.ServiceConfigurationError: com.fasterxml.jackson.databind.Module: Provider com.fasterxml.jackson.module.scala.DefaultScalaModule could not be instantiated"` and `Caused by: java.lang.NoSuchMethodError: scala.reflect.package$.classTag(Lscala/reflect/ClassTag;)Lscala/reflect/ClassTag;`. I can try to figure out what's going on there next week.

that said, let me know if you have any feedback on my approach here. I'd also love to get your thoughts on testing this. as far as I can tell, the `diff*` functions were not tested, I imagine because the logic was relatively simple and testing would require some mocking. there's a bit more logic here so could be worth implementing test. what do you think?

idreeskhan:
This can be closed now that the other one is merged, right?

jrmcglynn:
Yes, closing. #489 can be used for selecting partitions of natively-partitioned BQ tables.

PR #515, opened 2022-01-18
Add GenTestUtils

kellen:
Provides a standard way to log the seed on test failure and to handle generation failure.

### DIFF
--- a/ratatool-scalacheck/README.md
+++ b/ratatool-scalacheck/README.md
@@ -39,7 +39,9 @@ Implicit Arbitrary instances are also available for Avro and Protobuf records. E
 val avroArb: Arbitrary[MyRecord] = implicitly[Arbitrary[MyRecord]]
 ```
 
-Multiple `Gen` instances can be composed together with a for-comprehension. This will cause the data to be generated with the same seed.
+Multiple `Gen` instances can be composed together with a for-comprehension, which will cause the data to be generated with the same seed,
+which can aid in test reproducibility. See `withGen` below for a convenience function for capturing failing seeds.
+
 ```scala
 val genData: Gen[(MyRecord, Int)] = for {
  record <- avroGen
@@ -50,16 +52,9 @@ val genData: Gen[(MyRecord, Int)] = for {
 val data: Option[(MyRecord, Int)] = genData.sample
 ```
 
-Due to scalacheck internals generation can fail so the return type of `.sample` is `Option[A]`.
-In general, calling `.get` on an `Option` is an anti-pattern because this can produce `NoSuchElementException`.
-When used in conjunction with scalacheck, calling `.sample.get` can be the root cause of flaky tests.
-
-Repeatedly calling `.sample` is also an anti-pattern as each generated instance will be produced with a different random seed and tests will not be reproducible.
-Instead, use `withGen` as described below.
-
 ## In tests
 
-When using `Gen` to create data in a test, use `withGen` to capture the random seed and print it on failure:
+When using `Gen` to create data in a test, `withGen` can be used to capture the random seed and print it on failure:
 ```scala
 import com.spotify.ratatool.scalacheck.GenTestUtils
 
--- a/ratatool-scalacheck/README.md
+++ b/ratatool-scalacheck/README.md
@@ -9,9 +9,7 @@ Ratatool-Scalacheck contains classes and functions which help with using Scalach
 import com.spotify.ratatool.scalacheck._
 import org.scalacheck.Gen
 
-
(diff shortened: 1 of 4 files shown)


ratatool-scalacheck/README.md
```diff
@@ -39,7 +37,50 @@ Implicit Arbitrary instances are also available for Avro and Protobuf records. E
...
+Due to scalacheck internals generation can fail so the return type of `.sample` is `Option[A]`.
+In general, calling `.get` on an `Option` is an anti-pattern because this can produce `NoSuchElementException`.
+When used in conjunction with scalacheck, calling `.sample.get` can be the root cause of flaky tests.
+
+Repeatedly calling `.sample` is also an anti-pattern as each generated instance will be produced with a different random seed and tests will not be reproducible.
+Instead, use `withGen` as described below.
 
+## In tests
+
+When using `Gen` to create data in a test, use `withGen` to capture the random seed and print it on failure:
```
idreeskhan, line 62: Again I think we can reword this to be less suggestive. We don't want to tell users _what_ to do, just _how_ to do different things and enable them to make their own decisions

ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/GenTestUtils.scala
```diff
@@ -0,0 +1,76 @@
...
+   *   The result of `fn` applied to the generated `A`, or `None` on generation failure.
+   */
+  def withGen[A, T](genA: Gen[A], seed: Seed)(
+    fn: A => T
+  )(implicit line: sourcecode.Line, name: sourcecode.FileName): Option[T] = {
+    genA.apply(Gen.Parameters.default, seed) match {
+      case None =>
+        logger.error(
+          s"Failed to generate a valid value at ${name.value}:${line.value}. " +
+            s"Consider rewriting Gen instances to be less failure-prone. " +
```
idreeskhan, line 65: This is also kind of preachy. Should be something like `this can happen when XYZ`.

idreeskhan (commented):
I think I'm mostly fine with this being here, but I disagree with the way the documentation and messaging is written. We should focus on just conveying information to the user, some of the docs read like personal opinions which we want to avoid.

ratatool-scalacheck/README.md
```diff
@@ -39,7 +37,50 @@ Implicit Arbitrary instances are also available for Avro and Protobuf records. E
...
+ record <- avroGen
+ i <- Gen.choose(0, 10)
+} yield (record, i)
+
+// .sample gets a random seed
+val data: Option[(MyRecord, Int)] = genData.sample
+```
+
+Due to scalacheck internals generation can fail so the return type of `.sample` is `Option[A]`.
+In general, calling `.get` on an `Option` is an anti-pattern because this can produce `NoSuchElementException`.
```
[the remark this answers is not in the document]
kellen: This prescriptive tone was motivated by the substantial misuse of `Gen` we've seen internally, but I can definitely tone it down.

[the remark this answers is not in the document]
idreeskhan: I'm not sure I wholly agree with your definition of `misuse`. In some cases yes, but other times people do things like `.sample.get` because it's easier and they don't need such strong guarantees. There's nothing explicitly wrong with that, we just need to make sure they are capable of making an informed decision.

kellen:
@idreeskhan changed to a more neutral presentation, re-review?

ratatool-scalacheck/README.md
```diff
@@ -39,7 +37,45 @@ Implicit Arbitrary instances are also available for Avro and Protobuf records. E
 
 ```scala
 val avroArb: Arbitrary[MyRecord] = implicitly[Arbitrary[MyRecord]]
+```
+
+Multiple `Gen` instances can be composed together with a for-comprehension, which will cause the data to be generated with the same seed,
```
idreeskhan, line 42: Gonna nitpick more sorry, the text reads as if composing gens always results in failure. we should move the failure text down to sample, and adjust it to say something like
```
To extract a single value from a generator you can just `.sample`, however this can result in failures in cases where the seed is unable to produce a valid value. See `withGen` etc etc etc`

idreeskhan: Also can you add a header to this section to separate the `amend()` stuff

idreeskhan (commented):
Just a nitpick left, looks good. Thanks for your contribution!

ratatool-scalacheck/README.md
kellen: Reworded.

PR #517, opened 2022-01-20
add flogger backend dependency

benkonz:
I was getting a 

```
No logging platforms found in while using google flogger logging ClassNotFoundException: com.google.common.flogger.backend.system.DefaultPlatform
```

error when running the tests locally in ratatool. It looks like we need to add a logging backend to ratatool in order to have flogger function properly. According to the flogger docs, https://github.com/google/flogger#1-add-the-dependencies-on-flogger, we need to add https://mvnrepository.com/artifact/com.google.flogger/flogger-system-backend/0.7.4 to build.sbt.

I set the scope to `test` in order to avoid adding dependencies for our users binary.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -181,8 +181,7 @@ lazy val ratatoolSampling = project
       "com.twitter" %% "algebird-core" % algebirdVersion,
       "joda-time" % "joda-time" % jodaTimeVersion,
       "org.scalacheck" %% "scalacheck" % scalaCheckVersion,
-      "org.scalatest" %% "scalatest" % scalaTestVersion % "test",
-      "com.google.flogger" % "flogger-system-backend" % floggerVersion % "test"
+      "org.scalatest" %% "scalatest" % scalaTestVersion % "test"
     ),
     // In case of scalacheck failures print more info
     testOptions in Test += Tests.Argument(TestFrameworks.ScalaCheck, "-verbosity", "3"),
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -180,7 +181,8 @@ lazy val ratatoolSampling = project
       "com.twitter" %% "algebird-core" % algebirdVersion,
       "joda-time" % "joda-time" % jodaTimeVersion,
       "org.scalacheck" %% "scalacheck" % scalaCheckVersion,
-      "org.scalatest" %% "scalatest" % scalaTestVersion % "test"
+      "org.scalatest" %% "scalatest" % scalaTestVersion % "test",
+      "com.google.flogger" % "flogger-system-backend" % floggerVersion % "test"
```
idreeskhan, line 185: If the issues occurs in ParquetIOTest can we move this dep to ratatool-extras

idreeskhan (commented):
fine to merge this to unblock for now but the fact that I don't understand what is going on worries me

build.sbt
benkonz: The [docs](https://github.com/google/flogger#1-add-the-dependencies-on-flogger) for flogger say that "the dependency on flogger-system-backend is only required to be included when the binary is run", so I think there is a transitive dependency that is using flogger, but is excluding flogger-backend. Since we aren't supplying a logging backend, flogger is defaulting to throwing an exception.

The strange thing is that you weren't able to replicate it and the tests pass on CI.

idreeskhan:
Yea that's what I'm referring to. My gut says there's a better way to fix this than adding this dep

benkonz:
> Yea that's what I'm referring to. My gut says there's a better way to fix this than adding this dep

It looks like [these](https://github.com/spotify/ratatool/blob/master/ratatool-extras/src/test/scala/com/spotify/ratatool/io/ParquetIOTest.scala#L48) tests call `ParquetIO.writeToFile`, which calls [this](https://github.com/spotify/ratatool/blob/master/ratatool-extras/src/main/scala/com/spotify/ratatool/io/ParquetIO.scala#L77), which calls [this](https://github.com/spotify/ratatool/blob/master/ratatool-extras/src/main/java/com/spotify/ratatool/GcsConfiguration.java#L81), which has `private static final GoogleLogger logger = GoogleLogger.forEnclosingClass();`, I'm not sure of an alternative other than deleting the test files

PR #539, opened 2022-02-28
making coder explicit for row comparisons

skippaDaBitFlippa:
BigDiffy was inferring the wrong coder resulting in errors when NaN values were passed. This PR is explicitly defining the coder

### DIFF
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
@@ -360,9 +360,10 @@ object BigDiffy extends Command with Serializable {
     keyFn: TableRow => MultiKey,
     diffy: TableRowDiffy,
     ignoreNan: Boolean = false
-  )(implicit coder: Coder[TableRow]): BigDiffy[TableRow] = {
+  ): BigDiffy[TableRow] = {
     // replace quotation marks at the beginning or end of the argument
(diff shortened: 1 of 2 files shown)


ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -359,7 +360,7 @@ object BigDiffy extends Command with Serializable {
     keyFn: TableRow => MultiKey,
     diffy: TableRowDiffy,
     ignoreNan: Boolean = false
-  ): BigDiffy[TableRow] = {
+  )(implicit coder: Coder[TableRow]): BigDiffy[TableRow] = {
```
clairemcginty, line 363: you shouldn't have to summon an implicit coder here -- since you're being explicit about using `TableRowJsonCoder.of()` below :)

clairemcginty, line 363: if you want to avoid changing this API signature, you could remove this implicit coder from the signature and just add `implicit val trCoder = Coder.beam(TableRowJsonCoder.of())` to the method body! Or alternately, just do `diff(...)(ClassTag[TableRow], Coder.beam(TableRowJsonCoder.of())` 👍

skippaDaBitFlippa:
Done and tested 👍 thanks @clairemcginty !

PR #551, opened 2022-05-09
Support parquet in BigDiffy/BigSampler

clairemcginty:
this PR expands on the limited parquet-support in parquet-extras; moves all Parquet logic into ratatool-sampling and ratatool-diffy; and adds Parquet integration to BigDiffy/BigSampler/directSampler.

Note that Parquet can currently be written in Scio as case classes ("[parquet-typed](https://github.com/spotify/scio/blob/main/scio-parquet/src/main/scala/com/spotify/scio/parquet/types/syntax/SCollectionSyntax.scala#L35)"), or as Specific/GenericRecord Avro ("[parquet-avro](https://github.com/spotify/scio/blob/main/scio-parquet/src/main/scala/com/spotify/scio/parquet/avro/syntax/SCollectionSyntax.scala#L44)"). The code I added supports inputs of either type by reading all Parquet inputs as GenericRecords. This is possible because all Parquet files have schema information in the file metadata header; even if you write Parquet from case classes, it can still be converted to an Avro schema and read into GenericRecords.

This includes unit tests and I also tested all three commands with parquet-typed and parquet-avro data.


benkonz (approved):
LGTM!

benkonz:
@clairemcginty 
I ran BigDiffy on my hack days project with the Scio pipeline writing parquet files, and it worked! 
image

PR #564, opened 2022-07-18
Fix filepath wildcard for avro/parquet

steveroy0226:
Below changes have been tested and confirmed to work. Also proposing alternative approach since I'm not a big fan of guessing whether we are dealing with Avro vs Parquet:

Approach (1)
- This PR, minimal change. Basically checking to see if the matching files has the relevant extensions and routing it to either `AvroSampler` or `ParquetSample`

Approach (2) 
- Not in this PR, but I think we should create a new arg called `mode` (which can be either `parquet`, `avro` or `bigquery`) and have that be a required field. It will be a breaking change, but then it will help avoid guessing which Sampler to use.

### DIFF
--- a/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
+++ b/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
@@ -18,6 +18,7 @@ package com.spotify.ratatool.samplers
 
 import java.net.URI
 import java.nio.charset.Charset
+
 import com.google.api.services.bigquery.model.{TableFieldSchema, TableReference}
 import com.google.common.hash.{HashCode, Hasher, Hashing}
 import com.spotify.ratatool.samplers.util.SamplerSCollectionFunctions._
--- a/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
+++ b/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
@@ -18,7 +18,6 @@ package com.spotify.ratatool.samplers
 
 import java.net.URI
 import java.nio.charset.Charset
-
 import com.google.api.services.bigquery.model.{TableFieldSchema, TableReference}
 import com.google.common.hash.{HashCode, Hasher, Hashing}
 import com.spotify.ratatool.samplers.util.SamplerSCollectionFunctions._
@@ -34,10 +33,12 @@ import org.apache.avro.Schema
 import org.apache.avro.generic.GenericRecord
 import org.apache.beam.runners.dataflow.options.DataflowPipelineWorkerPoolOptions
 import org.apache.beam.sdk.io.FileSystems
+import org.apache.beam.sdk.io.fs.MatchResult.Metadata
 import org.apache.beam.sdk.io.gcp.bigquery.BigQueryHelpers
 import org.apache.beam.sdk.options.PipelineOptions
 import org.slf4j.LoggerFactory
 
+import scala.jdk.CollectionConverters._
 import scala.language.{existentials, higherKinds}
 import scala.util.Try
 import scala.reflect.ClassTag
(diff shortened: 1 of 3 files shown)


benkonz:
I'm a fan of adding an optional parameter called `--input-mode`, similar to how we do it in BigDiffy https://github.com/spotify/ratatool/tree/master/ratatool-diffy#usage, just to keep the names consistent. In general, I'm not a fan of "guessing" what the input type is, but I also don't want to break existing users' workflows, so adding an optional parameter that we use, and inferring the input type if that parameter isn't there would work for me!

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -253,8 +254,10 @@ object BigSampler extends Command {
       )
       // Prompts FileSystems to load service classes, otherwise fetching schema from non-local fails
       FileSystems.setDefaultPipelineOptions(opts)
+      val fileNames = FileSystems.`match`(input).metadata().asScala.map(_.resourceId().getFilename)
```
idreeskhan, line 257: This will incur a network call in cloud storage cases right? Do failures here produce meaningful errors?

scalastyle-config.xml
```diff
@@ -3,7 +3,7 @@
  <check level="error" class="org.scalastyle.file.FileTabChecker" enabled="true"></check>
  <check level="error" class="org.scalastyle.file.FileLengthChecker" enabled="true">
   <parameters>
-   <parameter name="maxFileLength"><![CDATA[800]]></parameter>
+   <parameter name="maxFileLength"><![CDATA[1000]]></parameter>
```
idreeskhan, line 6: Hmm, I think this scalastyle config is already aligned with other Spotify Scala projects. I don't think we should change this and should instead use ignore or refactor when absolutely necessary

steveroy0226: good point, reverting this and adding ignore file length check

steveroy0226:
> I'm a fan of adding an optional parameter called `--input-mode`, similar to how we do it in BigDiffy https://github.com/spotify/ratatool/tree/master/ratatool-diffy#usage, just to keep the names consistent. In general, I'm not a fan of "guessing" what the input type is, but I also don't want to break existing users' workflows, so adding an optional parameter that we use, and inferring the input type if that parameter isn't there would work for me!

@benkonz, I like the approach. I will create a ticket to add `--input-mode`. I agree that we shouldn't break existing users' workflows.

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -253,8 +259,10 @@ object BigSampler extends Command {
       )
       // Prompts FileSystems to load service classes, otherwise fetching schema from non-local fails
       FileSystems.setDefaultPipelineOptions(opts)
+      val fileNames = getFileMedata(input).map(_.resourceId().getFilename)
```
idreeskhan, line 262: Should be `Metadata`

idreeskhan (approved):
just the one typo

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
```diff
@@ -253,8 +259,10 @@ object BigSampler extends Command {
...
+
       input match {
-        case avroPath if input.endsWith("avro") =>
+        case avroPath if fileNames.exists(_.endsWith("avro")) =>
```
idreeskhan, line 265: can we log which one we resolve it as? might help us later on

ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala
steveroy0226: ah, changing getFileMedata to getMedata

Issue #178, opened 2019-11-20
avroOf does not support logical types


drobert:
I just started using ratatool today and am using it for CaseClass and Avro generation. I believe there's an issue in the way it handles, or fails to handle, logical types in avro.

I'm using java/scala type `BigDecimal`, which in schema looks like this (newlines added by me for easier reading):
```json
{
  "name":"cost",
  "type": {
    "type":"bytes",
    "logicalType":"decimal",
    "precision":10,
    "scale":2
  }
}
```
Decimal logical type is documented here: https://avro.apache.org/docs/current/spec.html#Logical+Types 

What's happening when attempting to use `avroOf[MyGenericRecord]` is the `type` pattern match (found [in AvroGenerator.scala](https://github.com/spotify/ratatool/blob/master/ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/AvroGenerator.scala#L163)) hits `Schema.Type.BYTES` and can generate `byte[0]`. 

Later, the AvroCodec, which *does* honor the BigDecimal logical type, attempts to convert `byte[0]` into a `BigInteger` (see https://github.com/apache/avro/blob/master/lang/java/avro/src/main/java/org/apache/avro/Conversions.java#L84), which fails with:

>  NumberFormatException("Zero length BigInteger")

I feel there's a few options which could be done independently here (I'll PR as I'm able):
1. document the current limitation
1. add support for common, documented logical types to ratatool. likely these can be done in independent PRs

anish749:
I'm spending some time on this, and will try to create a PR late next week..!

rydenius:
I see a corresponding problem for `{ type: integer, "logicalType": "date" }` which gives a `java.lang.ClassCastException: java.lang.Integer cannot be cast to org.joda.time.LocalDate`.

ryanstull:
Yeah I ran into this issue as well, this would be super helpful to have.

aasokan:
Ran into the same issue, is there a workaround for this?

RustedBones:
Fixed by #691

PR #624, opened 2023-03-14
Add logical decimal type support to avroOf

jamesmcm:
Attempt to add support for logical decimal types to avroOf See issue #178

It appears the Builder still fails when nested records rely on Avro Conversions since it only includes them in the generated classes for the nested record, and not for the parent record builder.

I also could not get it to work for nullable logical decimal types as the type coercion fails for the null array (it's a bit related to the above issue that we'd want it to call the Conversion and also handle nulls - e.g. 0 byte array -> null).

Note I also don't handle the case of a fixed underlying byte array size in the Avro schema, nor checking that the generated integer will fit inside the decimal's precision and scale - but at the moment the above issues are more blocking.

### DIFF
--- a/ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/AvroGenerator.scala
+++ b/ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/AvroGenerator.scala
@@ -169,22 +169,17 @@ trait AvroGeneratorOps {
           )
           .map(AvroValue(_))
 
-      case Schema.Type.BYTES => {
-        if (schema.getLogicalType.isInstanceOf[LogicalTypes.Decimal]) {
-          Arbitrary.arbInt.arbitrary.map(
-            // TODO: Check integer against decimal scale
-            // TODO: Handle fixed array size in Avro schema
-            i => AvroValue(ByteBuffer.wrap(BigInteger.valueOf(i).toByteArray))
-          )
-        } else {
-          boundedLengthGen.flatMap(n =>
-            Gen.listOfN(n, Arbitrary.arbByte.arbitrary).map { values =>
-              val bb = ByteBuffer.wrap(values.toArray)
-              AvroValue(bb)
-            }
-          )
+      case Schema.Type.BYTES =>
+        Option(schema.getLogicalType) match {
+          case Some(dt: LogicalTypes.Decimal) =>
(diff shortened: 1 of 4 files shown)


ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/AvroGenerator.scala
```diff
@@ -169,13 +169,22 @@ trait AvroGeneratorOps {
...
+          )
+        } else {
+          boundedLengthGen.flatMap(n =>
+            Gen.listOfN(n, Arbitrary.arbByte.arbitrary).map { values =>
+              val bb = ByteBuffer.wrap(values.toArray)
+              AvroValue(bb)
+            }
+          )
+        }
+      }
```
RustedBones, line 187:
```suggestion
      case Schema.Type.BYTES =>
        Option(schema.getLogicalType) match {
          case Some(dt: LogicalTypes.Decimal) =>
            val max = BigInt(10).pow(dt.getPrecision) - 1
            Gen.choose(-max, max).map(_.toByteArray).map(AvroValue.apply)
          case _ =>
            for {
              n <- boundedLengthGen
              bs <- Gen.listOfN(n, Arbitrary.arbByte.arbitrary)
            } yield AvroValue(ByteBuffer.wrap(bs.toArray))
        }
```

RustedBones:
The test are failing due to a bug in avro 1.8.2, where the generated code does not include the proper conversions requires to support logical types. I ran tests with avro 1.11.1 and they pass.

benkonz:
> The test are failing due to a bug in avro 1.8.2, where the generated code does not include the proper conversions requires to support logical types. I ran tests with avro 1.11.1 and they pass.

I believe we are tied to avro 1.8.2 because of [scio](https://github.com/spotify/scio/blob/main/build.sbt#L37)'s avro version, is there a workaround to get logical types working in avro 1.8.2?

RustedBones:
Done

benkonz (approved):
looks good to me! thanks

jamesmcm:
Thanks @RustedBones !

Will that be patched in Avro btw? As it was an issue I hit when working with `Option[BigDecimal]` like this too. It'd be nice to test that once it's fixed on the Avro side.

RustedBones:
It is fixed in avro since 1.9 (May 7, 2019). So if you use any of the newer version in your project, the logical-type conversions will work for you.

PR #664, opened 2023-09-08
Upgrade scio 0 13

catherinejelder:


.scala-steward.conf
```diff
@@ -2,7 +2,6 @@ updates.ignore = [
     { groupId = "com.google.apis", artifactId="google-api-services-bigquery" },
     { groupId = "org.apache.avro", artifactId="avro" },
     { groupId = "com.google.guava", artifactId="guava" },
-    { groupId = "com.spotify" },
```
catherinejelder, line 5: I suggest removing this so we get scio upgrade prs, even though we'll still have to manually check our build.sbt file for other dependency conflicts before merging them https://spotify.slack.com/archives/C0QC1CLGK/p1694198975357739?thread_ts=1694190469.037639&cid=C0QC1CLGK

catherinejelder:
it looks like codecov is configured to fail the build if coverage decreases at all. I'd like to change this but it's not straightforward (we don't seem to even have a codecov config right now) so I'd like to defer that till later.

benkonz (approved):
approved, @catherinejelder could you make a ticket to look into updating the codecov config so that these version bumps don't fail the builds?

steveroy0226 (approved):
lgtm

PR #676, opened 2023-10-06
downgrade some libs to match scio

catherinejelder:
it seems good to match versions for dependencies we share with scio


steveroy0226 (approved):
lgtm

PR #686, opened 2023-11-21
match scio deps

catherinejelder:
match dependencies from the scio bump that I just merged in separately.

the only significant change is that [some beam avro dependencies have moved](https://github.com/apache/beam/releases/tag/v2.52.0)


catherinejelder:
overtaken by https://github.com/spotify/ratatool/pull/689

PR #691, opened 2023-11-28
Improve avro support

RustedBones:
- Do not rely on beam serialization to generate avro records
- Add proper support of logical-types according to avro spec
- Refactor AvroBigDiffy to leverage type information

~Compile with avro 1.8 and run test with latest avro (avro 1.8 generates broken code)~

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -10,5 +10,5 @@ addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.11.1")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.2")
 
 libraryDependencies ++= Seq(
-  "org.apache.avro" % "avro-compiler" % "1.8.2"
+  "org.apache.avro" % "avro-compiler" % "1.11.3"
 )
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -9,6 +9,4 @@ addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.10.0")
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.11.1")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.2")
 
-libraryDependencies ++= Seq(
-  "org.apache.avro" % "avro-compiler" % "1.11.3"
-)
+libraryDependencies ++= Seq("org.apache.avro" % "avro-compiler" % "1.8.2")
--- a/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
+++ b/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
@@ -18,118 +18,149 @@
 package com.spotify.ratatool.diffy
 
 import com.spotify.scio.coders.Coder
-import org.apache.avro.{Schema, SchemaValidatorBuilder}
-import org.apache.avro.generic.GenericRecord
-import scala.util.Try
+import org.apache.avro.Schema
+import org.apache.avro.generic.{GenericData, GenericRecord, IndexedRecord}
 
 import scala.jdk.CollectionConverters._
 
 /** Field level diff tool for Avro records. */
-class AvroDiffy[T <: GenericRecord: Coder](
+class AvroDiffy[T <: IndexedRecord: Coder](
   ignore: Set[String] = Set.empty,
   unordered: Set[String] = Set.empty,
   unorderedFieldKeys: Map[String, String] = Map()
 ) extends Diffy[T](ignore, unordered, unorderedFieldKeys) {
 
-  override def apply(x: T, y: T): Seq[Delta] = {
-    new SchemaValidatorBuilder().canReadStrategy
-      .validateLatest()
-      .validate(y.getSchema, List(x.getSchema).asJava)
-    diff(Option(x), Option(y), "")
+  override def apply(x: T, y: T): Seq[Delta] = (x, y) match {
+    case (null, null)                    => Seq.empty
+    case (_, null)                       => Seq(Delta("", Some(x), None, UnknownDelta))
+    case (null, _)                       => Seq(Delta("", None, Some(y), UnknownDelta))
+    case _ if x.getSchema != y.getSchema => Seq(Delta("", Some(x), Some(y), UnknownDelta))
+    case _                               => diff(x, y, x.getSchema, "")
   }
 
-  def isAvroRecordType(schema: Schema): Boolean =
-    Schema.Type.RECORD.equals(schema.getType) ||
-      (Schema.Type.UNION.equals(schema.getType) &&
-        schema.getTypes.asScala.map(_.getType).contains(Schema.Type.RECORD))
-
-  private def diff(x: Option[GenericRecord], y: Option[GenericRecord], root: String): Seq[Delta] = {
-    // If a y exists we assume it has the superset of all fields, since x must be backwards
-    // compatible with it based on the SchemaValidator check in apply()
-    val schemaFields = (x, y) match {
-      case (Some(xVal), None) => xVal.getSchema.getFields.asScala.toList
-      case (_, Some(yVal))    => yVal.getSchema.getFields.asScala.toList
-      case _                  => List()
-    }
-
-    schemaFields
-      .flatMap { f =>
-        val name = f.name()
-        val fullName = if (root.isEmpty) name else root + "." + name
-        getRawType(f.schema()).getType match {
-          case Schema.Type.RECORD =>
-            val a = x.flatMap(r => Option(r.get(name).asInstanceOf[GenericRecord]))
-            val b = y.flatMap(r => Option(r.get(name).asInstanceOf[GenericRecord]))
-            (a, b) match {
-              case (None, None)       => Nil
-              case (Some(_), None)    => Seq(Delta(fullName, a, None, UnknownDelta))
-              case (None, Some(_))    => Seq(Delta(fullName, None, b, UnknownDelta))
-              case (Some(_), Some(_)) => diff(a, b, fullName)
-            }
-          case Schema.Type.ARRAY if unordered.contains(fullName) =>
-            if (
-              unorderedFieldKeys.contains(fullName)
-              && isAvroRecordType(f.schema().getElementType)
-            ) {
-              val l = x
-                .flatMap(outer =>
-                  Option(outer.get(name).asInstanceOf[java.util.List[GenericRecord]].asScala.toList)
-                )
-                .getOrElse(List())
-                .flatMap(inner =>
-                  Try(inner.get(unorderedFieldKeys(fullName))).toOption.map(k => (k, inner))
-                )
-                .toMap
-              val r = y
-                .flatMap(outer =>
-                  Option(outer.get(name).asInstanceOf[java.util.List[GenericRecord]].asScala.toList)
-                )
-                .getOrElse(List())
-                .flatMap(inner =>
-                  Try(inner.get(unorderedFieldKeys(fullName))).toOption.map(k => (k, inner))
-                )
-                .toMap
-              (l.keySet ++ r.keySet).flatMap(k => diff(l.get(k), r.get(k), fullName)).toList
-            } else {
-              val a = x
-                .flatMap(r => Option(r.get(name).asInstanceOf[java.util.List[GenericRecord]]))
-                .map(sortList)
-              val b = y
-                .flatMap(r => Option(r.get(name).asInstanceOf[java.util.List[GenericRecord]]))
-                .map(sortList)
-              if (a == b) {
-                Nil
-              } else {
-                Seq(Delta(fullName, a, b, delta(a.orNull, b.orNull)))
-              }
-            }
-          case _ =>
-            val a = x.flatMap(r => Option(r.get(name)))
-            val b = y.flatMap(r => Option(r.get(name)))
-            if (a == b) Nil else Seq(Delta(fullName, a, b, delta(a.orNull, b.orNull)))
-        }
-      }
-      .filter(d => !ignore.contains(d.field))
+  private def isRecord(schema: Schema): Boolean = schema.getType match {
+    case Schema.Type.RECORD => true
+    case Schema.Type.UNION  => schema.getTypes.asScala.map(_.getType).contains(Schema.Type.RECORD)
+    case _                  => false
   }
 
-  private def getRawType(schema: Schema): Schema = {
-    schema.getType match {
+  private def isNumericType(`type`: Schema.Type): Boolean = `type` match {
+    case Schema.Type.INT | Schema.Type.LONG | Schema.Type.FLOAT | Schema.Type.DOUBLE => true
+    case _                                                                           => false
+  }
+
+  private def numericValue(value: AnyRef): Double = value match {
+    case i: java.lang.Integer => i.toDouble
+    case l: java.lang.Long    => l.toDouble
+    case f: java.lang.Float   => f.toDouble
+    case d: java.lang.Double  => d
+    case _ => throw new IllegalArgumentException(s"Unsupported numeric type: ${value.getClass}")
+  }
+
+  private def diff(x: AnyRef, y: AnyRef, schema: Schema, field: String): Seq[Delta] = {
+    val deltas = schema.getType match {
       case Schema.Type.UNION =>
-        val types = schema.getTypes
-        if (types.size == 2) {
-          if (types.get(0).getType == Schema.Type.NULL) {
-            types.get(1)
-          } else if (types.get(1).getType == Schema.Type.NULL) {
-            // incorrect use of Avro "nullable" but happens
-            types.get(0)
-          } else {
-            schema
-          }
+        val data = GenericData.get()
+        val xTypeIndex = data.resolveUnion(schema, x)
+        val yTypeIndex = data.resolveUnion(schema, y)
+        if (xTypeIndex != yTypeIndex) {
+          // Use Option as x or y can be null
+          Seq(Delta(field, Option(x), Option(y), UnknownDelta))
         } else {
-          schema
+          // same fields, refined schema
+          val fieldSchema = schema.getTypes.get(xTypeIndex)
+          diff(x, y, fieldSchema, field)
         }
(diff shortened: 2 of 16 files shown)


ratatool-diffy/src/test/scala/com/spotify/ratatool/diffy/AvroDiffyTest.scala
```diff
@@ -200,70 +199,7 @@ class AvroDiffyTest extends AnyFlatSpec with Matchers {
     )
 
     du(x, y) should equal(
-      Seq(
-        Delta("repeated_record.nested_repeated_field", Option(jl(10, 20, 30)), None, UnknownDelta),
-        Delta("repeated_record.string_field", Option("b"), None, UnknownDelta)
```
RustedBones, line 205: This result does not make sense to me. We as we are keying by field, we should get the same output as map comparison.
I propose an output that puts the key in the field path.

ratatool-scalacheck/src/main/scala/com/spotify/ratatool/scalacheck/HashMapBuildable.scala
```diff
@@ -0,0 +1,32 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.spotify.ratatool.scalacheck
+
+import org.scalacheck.util.Buildable
+
+private[scalacheck] object HashMapBuildable {
```
RustedBones, line 21: Trying to port that upstream in https://github.com/typelevel/scalacheck/pull/1023

benkonz:
Hi! I'm taking a look! Thanks for the contribution!

project/plugins.sbt
```diff
@@ -10,5 +10,5 @@ addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.11.1")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.5.2")
 
 libraryDependencies ++= Seq(
-  "org.apache.avro" % "avro-compiler" % "1.8.2"
+  "org.apache.avro" % "avro-compiler" % "1.11.3"
```
catherinejelder, line 13: it makes me a bit nervous to be compiling and testing with two different versions of avro (if I'm correct that that's what happening here). is there anything preventing us from moving to avro "1.11.3" everywhere (I would guess we're blocked on something in scio)?

RustedBones:
Updated to latest scio. Code now also works for avro 1.8.2

clairemcginty (approved):

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
```diff
@@ -18,118 +18,164 @@
...
+    case i: java.lang.Integer => i.toDouble
+    case l: java.lang.Long    => l.toDouble
+    case f: java.lang.Float   => f.toDouble
+    case d: java.lang.Double  => d
+    case _ => throw new IllegalArgumentException(s"Unsupported numeric type: ${value.getClass}")
   }
 
-  private def getRawType(schema: Schema): Schema = {
-    schema.getType match {
+  private def diff(x: AnyRef, y: AnyRef, schema: Schema, field: String): Seq[Delta] = {
```
benkonz, line 85: would it be possible to get more specific types here than `AnyRef` for x and y?

RustedBones: I don't think this is possible.
The goal here is to call this recursively and branch the diff according to the schema type information + casting.

We can potentially create dedicated diff methods like `diffRecord(x: IndexedRecord, y: IndexedRecord)` but this does not play well with parameterized types like `list` and `maps`

benkonz: ah, I see. I trust your judgement, and any type other than `AnyRef` we choose probably would have to be super generic anyway.

ratatool-scalacheck/src/test/scala/com/spotify/ratatool/scalacheck/AvroGeneratorTest.scala
```diff
@@ -46,29 +40,25 @@ object AvroGeneratorTest extends Properties("AvroGenerator") {
...
     )
 
-  val richTupGen = (specificRecordOf[TestRecord], specificRecordOf[TestRecord]).tupled
-    .amend2(specificRecordOf[RequiredNestedRecord])(_.setRequiredFields, _.setRequiredFields)
+  val genRichTup = (genTestRecord, genTestRecord).tupled
+    .amend2(genRequiredNestedRecord)(_.setRequiredFields, _.setRequiredFields)
 
-  property("support RichAvroGen") = forAll(richGen) { r =>
+  property("support RichAvroGen") = forAll(genRich) { r =>
+    // format: off
```
monzalo14, line 47: Any reason to exclude this from formatting?

monzalo14 (commented):
Thanks for the contribution! I'd like to get a couple things clarified before approving. For some of the changes here, it's a bit hard to understand whether these could break any user code or not, without having full context on the recent avro/coder updates. Clarifying that to the extent possible would be ideal 👍 

On another topic, did you use a different formatter version than [what we currently have in prod](https://github.com/spotify/ratatool/blob/a42c883c5a6b8465a5b2ae1f7d8a56545e5356a4/.scalafmt.conf)? If possible, I'd like to reduce formatting changes to a minimum to keep better track of this, since it is a somewhat big PR.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
```diff
@@ -18,118 +18,164 @@
...
+class AvroDiffy[T <: IndexedRecord: Coder](
   ignore: Set[String] = Set.empty,
   unordered: Set[String] = Set.empty,
   unorderedFieldKeys: Map[String, String] = Map()
 ) extends Diffy[T](ignore, unordered, unorderedFieldKeys) {
 
-  override def apply(x: T, y: T): Seq[Delta] = {
-    new SchemaValidatorBuilder().canReadStrategy
-      .validateLatest()
-      .validate(y.getSchema, List(x.getSchema).asJava)
```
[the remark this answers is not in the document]
RustedBones: In terms of diff, avro schemas must be strictly equal. This is checked line 62 in the new version.

Compatible schema are used on read time to adapt stored data to the desired read schema. Once in memory, we should not compare data constructed with different schema, even if compatible.

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
```diff
@@ -18,118 +18,164 @@
...
-    diff(Option(x), Option(y), "")
-  }
+  private lazy val avroRuntimeVersion =
+    Option(classOf[Schema].getPackage.getImplementationVersion)
+
+  // after avro 1.8, use SpecificData.getForClass
+  private def dataForClass(cls: Class[_]): SpecificData = Try {
```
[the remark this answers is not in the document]
RustedBones: This is unused indeed. will remove. default specific data instance is enough to resolve type union

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
```diff
@@ -18,118 +18,164 @@
...
+          // Use Option as x or y can be null
+          Seq(Delta(field, Option(x), Option(y), UnknownDelta))
+        } else {
+          // same fields, refined schema
+          val fieldSchema = schema.getTypes.get(xTypeIndex)
+          diff(x, y, fieldSchema, field)
+        }
+      case Schema.Type.RECORD =>
+        val a = x.asInstanceOf[IndexedRecord]
+        val b = y.asInstanceOf[IndexedRecord]
```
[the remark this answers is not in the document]
RustedBones: if schema type is a record, `IndexedRecord` is the least powerful abstraction we need to check equality. We were previously using `GenericRecord` that extends `IndexedRecord`, but equality can be done on field order.

ratatool-diffy/src/test/scala/com/spotify/ratatool/diffy/AvroDiffyTest.scala
```diff
@@ -17,22 +17,20 @@
...
 import org.apache.beam.sdk.util.CoderUtils
-import org.scalacheck.Arbitrary
-
-import scala.jdk.CollectionConverters._
+import org.apache.beam.vendor.guava.v32_1_2_jre.com.google.common.io.BaseEncoding
 import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 
+import java.nio.ByteBuffer
+import scala.jdk.CollectionConverters._
```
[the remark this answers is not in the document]
RustedBones: it is prob my intelliJ organize import. I don't think we have a scalafmt rule for import order.

ratatool-diffy/src/test/scala/com/spotify/ratatool/diffy/AvroDiffyTest.scala
```diff
@@ -209,72 +207,7 @@ class AvroDiffyTest extends AnyFlatSpec with Matchers {
...
 
     du(x, y) should equal(
-      Seq(
-        Delta("repeated_record.nested_repeated_field", Option(jl(10, 20, 30)), None, UnknownDelta),
-        Delta("repeated_record.string_field", Option("b"), None, UnknownDelta)
-      )
-    )
-  }
-
-  it should "support schema evolution if ignored" in {
```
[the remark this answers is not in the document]
RustedBones: it is not covered. records with different schemas are not equal, as explained above.

ratatool-scalacheck/src/test/scala/com/spotify/ratatool/scalacheck/AvroGeneratorTest.scala
RustedBones: It looks nicer to have all test case as one liner even if this exceed max line with manually aligned custom `|:` operator

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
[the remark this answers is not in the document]
idreeskhan: Can you expand on what you mean? I'm a bit confused reading this thread

ratatool-diffy/src/test/scala/com/spotify/ratatool/diffy/AvroDiffyTest.scala
idreeskhan: I'd want to verify my understanding of how this output looks like with multiple nestings. I think if we are putting it on the field path we probably want to also recursively pass the field keys which I don't think is happening here https://github.com/spotify/ratatool/pull/691/files#diff-c02df8e7364f3ad968c28ac24f66eb4de3e9f15ef79dd7f05f5c05b1ecb98225R112

Though perhaps I'm misreading something here

RustedBones: The key is altered wen presenting the diff result here: https://github.com/spotify/ratatool/pull/691/files#diff-c02df8e7364f3ad968c28ac24f66eb4de3e9f15ef79dd7f05f5c05b1ecb98225R115

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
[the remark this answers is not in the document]
RustedBones: avro record with different models must not be compared.

schema compatibility is relevant when reading, making sure the `writerSchema` and the `readerSchema` are compatible. Once read, the records strictly folow the `readerSchema` where **field index** matters. Strict schema equality must be ensured before comparing content.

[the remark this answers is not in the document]
idreeskhan: Ok, trying to confirm my understanding here since I'm still a bit confused.
Dataset A is updated to add new nullable field `x` and becomes Dataset A'.
We go to diff these two datasets.
Are you saying that this field comparison will end up in one of the above null cases prior to the schema comparison?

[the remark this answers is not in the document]
RustedBones: If the new field has a default value, we should compare the 2 datasets using either A or A' schema when reading.
- using the A schema -> the new field won't be present in both side when reading, diff won't find any issue.
- using the A' schema -> we'll get the diffs where the value is present in the new dataset.
Obtained schema for those records are always equal

Now, if we read A with A schema and A' with A' schema, we should not even compare records.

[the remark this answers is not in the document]
idreeskhan: I would go one step further and say it should necessarily be the A' schema, even if it's nullable/has default, and cases where a field is missing should fail. IIRC this is the current functionality. It's unclear to me if/where this behaviour is retained

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
```diff
@@ -18,118 +18,167 @@
...
-      .filter(d => !ignore.contains(d.field))
+  private def isNumericType(`type`: Schema.Type): Boolean = `type` match {
+    case Schema.Type.INT | Schema.Type.LONG | Schema.Type.FLOAT | Schema.Type.DOUBLE => true
+    case _                                                                           => false
   }
 
-  private def getRawType(schema: Schema): Schema = {
-    schema.getType match {
+  private def numericValue(value: AnyRef): Double = value match {
+    case i: java.lang.Integer => i.toDouble
```
idreeskhan, line 64: Why `.toDouble` here?

RustedBones: Because [numericDelta](https://github.com/spotify/ratatool/blob/51ec7981aa283b6ca07742474166666780afbcdc/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/Diffy.scala#L118) only supports double

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/AvroDiffy.scala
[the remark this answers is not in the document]
RustedBones: The [BigDiffy](https://github.com/spotify/ratatool/blob/51ec7981aa283b6ca07742474166666780afbcdc/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala#L145) API of this lib is not file aware. It only works in terms of in-memory records and can't make any assumption on writer schema.

It is up to the users creating the `SCollection` to read using the correct schema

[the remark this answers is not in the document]
clairemcginty: > The [BigDiffy](https://github.com/spotify/ratatool/blob/51ec7981aa283b6ca07742474166666780afbcdc/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala#L145) API of this lib is not file aware.

cc @RustedBones, AFAIK, BigDiffy is file-aware when run through the CLI, which invokes BigDiffy#run: https://github.com/spotify/ratatool/blob/51ec7981aa283b6ca07742474166666780afbcdc/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala#L683-L760

but even then, it looks like it selects the [schema associated with the RHS](https://github.com/spotify/ratatool/blob/51ec7981aa283b6ca07742474166666780afbcdc/ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala#L751-L754) and uses that for both resulting SCollections. So maybe we could add schema validation there (ensure that RHS schema is equal to, or a superset of, the LHS schema)?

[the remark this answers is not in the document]
RustedBones: Added an extra check that will prefer backward compatible reader schema.
When schemas are different, but both forward and backward compatible, will print a warning

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
```diff
@@ -745,13 +748,30 @@ object BigDiffy extends Command with Serializable {
...
+        val rhsReader = SchemaCompatibility.checkReaderWriterCompatibility(rhsSchema, lhsSchema)
+
+        import SchemaCompatibility.SchemaCompatibilityType._
+        val schema = (lhsReader.getType, rhsReader.getType) match {
+          case (COMPATIBLE, COMPATIBLE) =>
+            if (lhsSchema != rhsSchema) {
+              logger.warn("Avro schemas are compatible, but not equal. Using schema from {}", rhs)
+            }
+            rhsSchema
+          case (COMPATIBLE, INCOMPATIBLE) =>
```
idreeskhan, line 767: This is a change in underlying functionality, IMO We should also warn in these cases rather than proceed transparently

idreeskhan (approved):

ratatool-diffy/src/main/scala/com/spotify/ratatool/diffy/BigDiffy.scala
idreeskhan: Honestly I still think we enforce RHS backwards compatibility unless otherwise shown to be necessary, but if we are changing functionality/flexibility then we need to do so in a way that is transparent to users.

I'll leave the actual decision here down to current members of the owning team

RustedBones: reverted to previous behavior using the `SchemaValidatorBuilder` that thows a `SchemaValidationException` with detailed error in case of incompatibility.

PR #696, opened 2024-01-22
Upgrade to Scio 0.14.0

clairemcginty:
Scio 0.14.0 upgrade


build.sbt
```diff
@@ -153,6 +159,7 @@ lazy val ratatoolCommon = project
       "org.apache.avro" % "avro" % avroVersion,
       "org.apache.avro" % "avro-mapred" % avroVersion classifier "hadoop2",
       "com.google.guava" % "guava" % guavaVersion,
+      "org.apache.beam" % "beam-vendor-guava-32_1_2-jre" % beamVendorVersion,
```
RustedBones, line 162: I don't think common should depend on beam. It is fine to use non shaded guava IMHO

monzalo14:
Thanks for the update @clairemcginty @RustedBones!

Issue #699, opened 2024-02-01
Improve sample population selection for deterministic hashing


idreeskhan:
Currently BigSampler tends to undersample for a given input ratio when performing a deterministic sample

One avenue to explore is:

Once hashes are created they are normalized in a [0.0, 1.0] range by [boundLong](https://github.com/spotify/ratatool/blob/ae43f56ea139d5dccab1e91b0b877e8ca7a39d9e/ratatool-sampling/src/main/scala/com/spotify/ratatool/samplers/BigSampler.scala#L75). Potentially this function should be updated or modified. One possible way is by the upper/lower bound of the input results, however this may be difficult to implement in practice. It could also be removed and replaced, the specifics of this implementation are lost to time and have dropped out of my memory.

Another path instead or in addition to this is:

We primarily use farmhash, which is not a cryptographic hash function. Is the output sufficiently uniform in its distribution? If not, now that additional hashes are available within BigQuery, is there another function with a more appropriate output distribution

PR #700, opened 2024-02-01
Add docs for reproducing sample from BigQuery

idreeskhan:


ratatool-sampling/README.md
```diff
@@ -85,6 +85,16 @@ Leveraging `--fields=<field1,field2,...>` BigSampler can produce a hash based on
...
  of 50% of users. If multiple records contain the same `user_id` they will all be in or out of the
  sample.
+
+### Reproducing within BigQuery
+Currently, BigSampler defaults to Farmhash, which is also used in BigQuery. When sampling with a seed and one or more fields,
+ under the hood Farmhash will create a byte array, convert all inputs to bytes, and concatenate them together. To recreate this in BigQuery, you
+ will have to pre-create the seed as a little endian hex encoded byte string, as BigQuery does not currently allow directly converting an integer
+ to bytes.
+
+`FARM_FINGERPRINT(CONCAT(b'\x2A\x00\x00\x00', b'abc'))` will produce the equivalent hash of `--seed=42` with one `fields` where the given record has value `abc`.
```
RickardZwahlen, line 95:
```suggestion
`FARM_FINGERPRINT(CONCAT(b'\x2A\x00\x00\x00', CAST('abc' as BYTES))` will produce the equivalent hash of `--seed=42` with one `fields` where the given record has value `abc`.
```

Since most users won't want to hash a static String, wouldn't this syntax be more useful?

ratatool-sampling/README.md
```diff
@@ -85,6 +85,16 @@ Leveraging `--fields=<field1,field2,...>` BigSampler can produce a hash based on
  are in the sample. For example, `--fields=user_id --sample=0.5` will always produce the same sample
  of 50% of users. If multiple records contain the same `user_id` they will all be in or out of the
  sample.
+
+### Reproducing within BigQuery
+Currently, BigSampler defaults to Farmhash, which is also used in BigQuery. When sampling with a seed and one or more fields,
```
benkonz, line 90: could you link to [Farmhash](https://github.com/google/farmhash)? Also, it looks like the FarmHash repo has been archived. Considering it's used in BigQuery, it shouldn't be much of an issue, but is it possible that we should explore an alternate hashing algo in the future?

idreeskhan: I think if we do it would be for the distribution work in #699 , or if farmhash stops working. ATM it's still supported through BQ so not super worried

idreeskhan: added a link earlier

PR #902, opened 2026-07-24
Fix BigDiffy NPE on parquet schema mismatch in array fields

monzalo14:
## Problem

When running BigDiffy in parquet mode, parquet can produce null values for non-nullable array/map fields. This happens both when:
1. The reader schema has fields missing from the writer (schema evolution)
2. The parquet data itself contains nulls in non-nullable columns

In either case, Avro's `GenericDatumWriter.getArraySize` calls `((Collection) null).size()` with no null guard, causing a `NullPointerException`.

`--nullableCoders` does not help -- that flag operates at the Scio coder boundary, not inside Avro record serialization.

## Solution

After selecting the compatible schema in `getCompatibleSchemaForFiles`, wrap ALL non-nullable array and map fields in a union with null via `makeCollectionFieldsNullable`. The original type is placed first in the union to preserve existing default values (e.g., `[]` for arrays), and null is added second so the Avro coder can serialize null values without NPE. The transformation recurses into nested record fields.

Also adds an early return when both schemas are identical and no collection fields need fixing.

## Test plan

- 3 unit tests for `makeCollectionFieldsNullable` in `ParquetIOTest`: arrays+maps, already-nullable fields, no-collection-fields
- 1 integration test in `BigDiffyTest`: writes parquet files with mismatched schemas, verifies the compatible schema wraps collection fields in nullable unions, and confirms AvroCoder can encode/decode records with null array fields
- Reproduced the exact NPE against the reporter's actual data, partition `2026-07-20T10`) using a local ratatool build; verified the NPE is gone with this fix
- All 29 existing + new tests pass on both 2.12 and 2.13


ratatool-diffy/src/test/scala/com/spotify/ratatool/diffy/BigDiffyTest.scala
```diff
@@ -445,6 +445,75 @@ class BigDiffyTest extends PipelineSpec {
...
+    // Simulate what the parquet reader produces for a record missing the tags column:
+    // a GenericRecord with the compatible schema where tags is null.
+    // Without the nullable union fix, encoding this record NPEs in
+    // GenericDatumWriter.getArraySize because it calls ((Collection) null).size().
+    val beamCoder = CoderMaterializer.beamWithDefault(avroGenericRecordCoder(compatSchema))
+    val simulatedRhsRecord = new GenericData.Record(compatSchema)
+    simulatedRhsRecord.put("id", 1)
+    simulatedRhsRecord.put("tags", null)
+
+    val encoded = CoderUtils.encodeToByteArray(beamCoder, simulatedRhsRecord)
```
jimmy-james: Blocking: this proves the transformed schema can encode a hand-built null record, but it never runs BigDiffy. I extended this fixture through `diffParquet` and `sc.run()`, and it still fails with an NPE while Kryo decodes the array stored in a `Delta`. Could we make this a full-pipeline regression and fix that downstream serialization path?

ratatool-sampling/src/main/scala/com/spotify/ratatool/io/ParquetIO.scala
```diff
@@ -72,8 +74,58 @@ object ParquetIO {
...
+        // null second so the coder tolerates nulls from the parquet reader.
+        val nullableType =
+          Schema.createUnion(field.schema(), Schema.create(Schema.Type.NULL))
+        new Schema.Field(
+          field.name(),
+          nullableType,
+          field.doc(),
+          field.defaultVal()
+        )
+      } else if (fieldType == Schema.Type.RECORD) {
```
jimmy-james: Blocking: recursion only handles bare `RECORD` fields. Arrays inside nullable record unions or array/map child records stay non-nullable; focused tests for both shapes still returned `ARRAY`. Could we recursively visit union branches, array elements, and map values, with cycle handling for named recursive records?

ratatool-sampling/src/main/scala/com/spotify/ratatool/io/ParquetIO.scala
```diff
@@ -72,8 +74,58 @@ object ParquetIO {
...
+    if (!hasFieldsToFix) return schema
+
+    val newFields = schema.getFields.asScala.map { field =>
+      val fieldType = field.schema().getType
+      if (collectionTypes.contains(fieldType) && !isNullableSchema(field.schema())) {
+        // Original type first so any existing default value stays valid;
+        // null second so the coder tolerates nulls from the parquet reader.
+        val nullableType =
+          Schema.createUnion(field.schema(), Schema.create(Schema.Type.NULL))
+        new Schema.Field(
```
jimmy-james: Blocking: these constructors drop field order, aliases, and custom properties, and the new record also loses its aliases and properties. A focused aliased-schema test lost the record alias immediately. Could we use the Avro field copy constructor and explicitly preserve the record metadata?

